Issue #110, opened 2020-03-22
Automatically detect bids with missing nonces


mslipper:
Currently, Bob will display an opaque error if users attempt to reveal bids with missing nonces. Rather than displaying an error, we should instead display a warning on the auction screen that gives users the option to import the nonce graphically.

mslipper:
For bids with missing nonces:

- Display a warning underneath the TLD with a link to an import bid modal
- Modal just displays an import bid field and a button that calls the import RPC call with the right address
- The 'reveal' button should say 'Nonce import needed' and display the modal until a valid nonce is inputted

mslipper:
Draft PR EOW

PR #116, opened 2020-03-25
Warn users when bid is missing (from rescan) and allow repair of missing bid value

pinheadmz:
Closes https://github.com/kyokan/bob-wallet/issues/110

### Changes summary

- Added `importNonce` and `getNonce` to wallet IPC calls
- Fixed a typo in "send" modal (placeholder only had 5 decimals)
- Auction page:
  - List of bids is shown (not collapsed) by default for all states
  - If a bid is "owned" but had no "value" it is broken...
    - UI shows a red warning instead of reveal button
    - Total bids shows `?` instead of `NaN`
    - Bid value in table shows alert emoji with link to repair bid
      - User clicks to repair, enters values until they enter the correct value
      - Each change in the input computes a blind from the given value and compares against the known blind
      - Once the blinds match, the input is disabled, and briefly turns green while the nonce is imported to the wallet, and namestate is reloaded.

### A healthy wallet in reveal phase:

Screen Shot 2020-03-25 at 1 49 18 PM

### A wallet after rescan, "forgot" the bid value:

Screen Shot 2020-03-25 at 1 50 55 PM

### User enters the bid value they remember or have backed up:

Mar-25-2020 14-28-09

### DIFF
--- a/app/background/wallet/service.js
+++ b/app/background/wallet/service.js
@@ -246,6 +246,15 @@ class WalletService {
     },
   );
 
+  getNonce = async (options) => {
+    this._ensureClient();
+    return this.client.getNonce(WALLET_ID, options.name, options);
+  };
+
+  importNonce = async (options) => {
+    return this._executeRPC('importnonce', [options.name, options.address, options.bid]);
+  }
+
   _onNodeStart = async (networkName, network, apiKey) => {
     this.networkName = networkName;
     const walletOptions = {
@@ -343,6 +352,8 @@ const methods = {
   lock: service.lock,
   unlock: service.unlock,
   isLocked: service.isLocked,
+  getNonce: service.getNonce,
+  importNonce: service.importNonce,
 };
 
 export async function start(server) {
--- a/app/pages/Auction/BidActionPanel/Reveal.js
+++ b/app/pages/Auction/BidActionPanel/Reveal.js
@@ -146,7 +146,7 @@ class Reveal extends Component {
     }
 
     if (needsRepair) {
-      text = 'Repair unknown bids';
+      text = 'Repair Unknown Bids';
       cname = 'domains__bid-now__action__warning';
     }
 
--- a/app/pages/Auction/BidActionPanel/Reveal.js
+++ b/app/pages/Auction/BidActionPanel/Reveal.js
@@ -114,7 +114,7 @@ class Reveal extends Component {
         <div className="domains__bid-now__title">Your Bids</div>
         <div className="domains__bid-now__content">
           <AuctionPanelHeaderRow label="Total Bids:">
-            {displayBalance(totalBids, true)}
+            {totalBids < 0 ? '?' : displayBalance(totalBids, true)}
           </AuctionPanelHeaderRow>
           <AuctionPanelHeaderRow label="Total Masks:">
             {displayBalance(totalMasks, true)}
@@ -131,9 +131,11 @@ class Reveal extends Component {
       hasRevealableBid,
       hasRevealed,
       hasPendingReveal,
+      needsRepair,
     } = this.props;
 
     let text = 'Reveal Your Bids';
+    let cname = 'domains__bid-now__action__cta'
 
     if (hasRevealed) {
       text = 'Bid Successfully Revealed';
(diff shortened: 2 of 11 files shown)


mslipper:
Lovely thanks, taking a look at this now.

app/background/wallet/service.js
```diff
@@ -246,6 +246,15 @@ class WalletService {
     },
   );
 
+  getNonce = async (options) => {
+    this._ensureClient();
+    return this.client.getNonce(WALLET_ID, options.name, options);
+  };
+
+  importNonce = async (options) => {
+    return this._executeRPC('importnonce', [options.name, options.address, options.bid]);
```
mslipper: Does this require a wallet to be selected?

app/pages/Auction/RepairBid.js
```diff
@@ -0,0 +1,85 @@
...
+    const value = val.match(/[0-9]*\.?[0-9]{0,6}/g)[0];
+    if (value * consensus.COIN > consensus.MAX_MONEY)
+      return;
+    this.setState({value: value});
+    this.verifyBid(value);
+  }
+
+  verifyBid(value) {
+    const {bid} = this.props;
+    walletClient.getNonce({
```
mslipper: Can you add a `catch` handler to this that surfaces errors using the error popup in use elsewhere?

mslipper:
Minor changes requested, otherwise LGTM. I prefer the way you've done it here to the modal UX I described earlier. Thank you! This will really help our users.

mslipper:
Quick question: Does this depend on the new `hsd` endpoints? If so we'll need the package.json to be updated to an `hsd` fork until the new endpoints land in master.

app/pages/Auction/BidActionPanel/Reveal.js
```diff
@@ -143,12 +145,17 @@ class Reveal extends Component {
       text = 'Reveal Pending...';
     }
 
+    if (needsRepair) {
+      text = 'Repair unknown bids';
```
mslipper: Please capitalize: `Repair Unknown Bids`.

app/background/wallet/service.js
pinheadmz: This is handled in `_executeRPC()`:

https://github.com/kyokan/bob-wallet/blob/b365bbaf52d984c7a138e53514f50a645cf85d71/app/background/wallet/service.js#L311-L314

https://github.com/kyokan/bob-wallet/blob/b365bbaf52d984c7a138e53514f50a645cf85d71/app/background/wallet/service.js#L280-L294

app/pages/Auction/RepairBid.js
pinheadmz: added in 24c02687b530c63fd366e293d86c704138cc66cd

app/pages/Auction/BidActionPanel/Reveal.js
pinheadmz: fixed in 1521a6672fad0dc28b9b0e3da014fb33ccf7b12e

pinheadmz:
@mslipper 

> Quick question: Does this depend on the new hsd endpoints? If so we'll need the package.json to be updated to an hsd fork until the new endpoints land in master.

It does not. However, we should update hsd anyway to include https://github.com/handshake-org/hsd/pull/387

app/pages/Auction/RepairBid.js
```diff
@@ -0,0 +1,96 @@
...
+      this.verifyBid(value);
+  }
+
+  verifyBid(value) {
+    const {bid} = this.props;
+    walletClient.getNonce({
+      name: bid.name,
+      address: bid.from,
+      bid: value * consensus.COIN
+    }).then(
```
pinheadmz: I think I can use async/await syntax here, I was confused about how to do that inside a react action but I'm pretty sure I can clean this up.

pinheadmz: OK, cleaned this up.

pinheadmz:
@mslipper cleaned up the async code around `importNonce` and `getNonce`. Also noticed the `NaN` was still present in the bid screen _before_ reveal period, so made the same fix there. Users can "repair bids" before the bidding phase is over.

mslipper:
LGTM, thanks!

PR #124, opened 2020-04-10
Display Bob version in settings tab and "About" panel

pinheadmz:
Closes https://github.com/kyokan/bob-wallet/issues/112

Also updated a bit of the menu syntax.

Screen Shot 2020-04-10 at 4 17 41 PM

### DIFF
--- a/app/menu.js
+++ b/app/menu.js
@@ -1,8 +1,28 @@
 import { app, Menu, shell } from 'electron';
 import showMainWindow, { getMainWindow } from './mainWindow';
+const pkg = require('../package.json');
+const Path = require('path');
 
 export default class MenuBuilder {
   buildMenu() {
+
+    // https://www.electronjs.org/docs/all#appsetaboutpaneloptionsoptions
+    // version: MacOS only
+    // credits: MacOS & Windows only
+    // authors: Linux only
+    // website: Linux only
+    // iconPath: Linux & Windows only
+    app.setAboutPanelOptions({
+      applicationName: pkg.productName, 
+      applicationVersion: pkg.version,
+      copyright: `License: ${pkg.license}`,
+      version: pkg.version,
+      credits: pkg.author,
+      author: pkg.author,
+      website: pkg.homepage,
+      iconPath: Path.join(__dirname, '../resources/icons/128x128.png')
+    });
+
     this.setupDevelopmentEnvironment();
 
     const template =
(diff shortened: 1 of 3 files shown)


app/menu.js
```diff
@@ -1,8 +1,28 @@
...
+    // iconPath: Linux & Windows only
+    app.setAboutPanelOptions({
+      applicationName: pkg.productName, 
+      applicationVersion: pkg.version,
+      copyright: `License: ${pkg.license}`,
+      version: pkg.version,
+      credits: pkg.author,
+      author: pkg.author,
+      website: pkg.homepage,
+      iconPath: Path.join(__dirname, '../resources/icons/128x128.png')
```
pinheadmz, line 23: @mslipper this may need to be tested on linux / windows specifically

mslipper:
Dig it!

app/menu.js
mslipper: I think you should use this guy, coupled with `path.join`: https://www.electronjs.org/docs/api/process#processresourcespath-readonly

app/menu.js
```diff
@@ -39,22 +59,22 @@ export default class MenuBuilder {
...
+          role: 'about'
         },
         {type: 'separator'},
         {label: 'Services', submenu: []},
         {type: 'separator'},
         {
           label: 'Hide Bob',
           accelerator: 'Command+H',
-          selector: 'hide:'
+          role: 'hide'
```
mslipper, line 70: Are the colons you removed here causing issues? Curious what these changes fix.

pinheadmz: We can remove this commit if you like. What happened was, I was reviewing the existing menu code and couldn't find documentation for the `selector` property. Then I found this: https://github.com/electron/electron/pull/3554 indicating that `selector` was deprecated and replaced by `role`. The available values for `role` are documented here: https://www.electronjs.org/docs/all#roles

mslipper: Nope, works for me. Glad this has been updated.

app/menu.js
pinheadmz: Fixed in ffd8470d8e27ab7079aade315aeb0e7354166b4b

This works in the MacOS build (although I think it is actually ignored by MacOS, see https://www.electronjs.org/docs/api/app#appsetaboutpaneloptionsoptions)

I spent a day trying and failing to get Windows installed on a partition on my MacBook Pro so I still can't currently test on Windows. Can you check this works? I'll acquire some dedicated hardware for Window soon.

mslipper: I'll check this out when we build the next release. Merging for now.

Issue #100, opened 2020-03-09
[Performance] Bids and Domains take a very long (>5mins) time to load


sdtsui:
This user owns a high number of tlds and is noting that it can take >5mins to go from this state:

![Screen Shot 2020-03-09 at 2 28 08 PM](https://user-images.githubusercontent.com/8230144/76259372-b5f6b880-6212-11ea-903d-d8542fe676ed.png)

to this state:
![Screen Shot 2020-03-09 at 2 29 51 PM](https://user-images.githubusercontent.com/8230144/76259366-b3945e80-6212-11ea-8dd8-d7235b7e17f4.png)

throughout all this, one can't look at their existing domains.
![Screen Shot 2020-03-09 at 2 29 38 PM](https://user-images.githubusercontent.com/8230144/76259369-b4c58b80-6212-11ea-9ca7-2d769b9e5ee0.png)

Can we speed this up by caching some data locally, or do some sort of progressive loading where individual UI elements are still accessible?  If the latter, a user can make changes even if bid status is not fully loaded for all names.

mslipper:
May be an issue with `hsd`.

pinheadmz:
I'm looking into this now. It might be nice to have pagination functions in the hsd wallet. There is some work in this area in bcoin that was never reviewed: https://github.com/bcoin-org/bcoin/pull/605

A work around for now might be to somehow abort API calls when the user clicks anything during a "loading" state. Might look into queuing up those calls for these views and allowing the queue processing to be cancelled.

pinheadmz:
...not to mention every single name in the bids table calls `getNameInfo()` twice: once in the `BidStatus` component and once again in the same table row on the `BidTimeLeft` component...

PR #126, opened 2020-04-15
Reduce API calls from certain elements by keeping global state up-to-date and caching redundant data

pinheadmz:
Closes https://github.com/kyokan/bob-wallet/issues/100

This PR reduces API calls in certain views to improve UI performance. The "Your Bids" view in particular has a table with three elements per bid: each of these elements was individually calling `getNameInfo()` on the same name! In addition, the use of `componentWillMount()` in certain elements to update state is insufficient since a new block event (and state update) will not trigger this method. The elements themselves should just rely on the global state being updated already.

Since bids and namestates only change when new blocks connect to the chain, we only really need to gather this data once per block. The data is already stored in the global state, so all we have to do is change _when_ it is updated.

This patch adds a new function `onNewBlock()` to `backgroundMonitor` which is executed any time a `doPoll()` encounters a new block from hsd. This function gets current fee rates, gets all bids, and then iterates through all the bids getting updated namestates. This means that on every new block, UI may be a bit sluggish for a minute. A new progress indicator is added in the sidebar footer.

Tested by running a bash script that opened 300 names and then bid once or twice on each name.

Future work:
- Add pagination to hsd wallet (see https://github.com/bcoin-org/bcoin/pull/605)
- Add functionality either in Bob or hsd to filter out expired auctions or bids that "don't matter anymore" for one reason or another.
- Switch the hsd<->Bob link from 10-second polling to **websocket events**. If possible, this will keep Bob up to date with essentially zero lag.
- Explore `getNameInfo` for ways to reduce 2 API calls (`node.getNameInfo` + `wallet.getAuctionInfo`) to one, or cache some data if possible.

### Demonstration

UI elements seem more responsive than before even when a new block is being processed. Once the new block processing is done, everything is perfectly snappy since no new API calls are made when loading/changing views.

Note the new "New Block Status" message just above the current height in the sidebar footer.

LoadBidsOnNewBlock

### DIFF
--- a/app/ducks/walletReducer.js
+++ b/app/ducks/walletReducer.js
@@ -25,7 +25,7 @@ export function getInitialState() {
       confirmed: '0',
       unconfirmed: '0'
     },
-    transactions: [],
+    transactions: new Map(),
     idle: 0,
     walletSync: false,
     walletSyncProgress: 0,
@@ -56,7 +56,7 @@ export default function walletReducer(state = getInitialState(), {type, payload}
           ...state.balance
         },
         isLocked: true,
-        transactions: []
+        transactions: new Map()
       };
     case UNLOCK_WALLET:
       return {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## Unreleased
 ### Fixed
 - Fixed total amount received by a transaction with multiple outputs to the wallet
+- Fixed sluggish UI from mass of redundant rpc calls on "Your Bids" view
 
 ### Added
 - Added warning for missing bid values and functionality to repair missing bids
--- a/app/background/wallet/service.js
+++ b/app/background/wallet/service.js
@@ -322,7 +322,7 @@ class WalletService {
   };
 
   async _executeRPC(method, args) {
-    this._selectWallet();
+    await this._selectWallet();
     return this.client.execute(method, args);
   }
 }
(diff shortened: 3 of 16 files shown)


pinheadmz:
rebase to 525a7e56f5a6d43a356d950b6b2e199bd5458758:
rebasing on master

mslipper:
https://www.youtube.com/watch?v=VI6dsMeABpU. Will review in the AM.

pinheadmz:
TODO: Use newBlockStatus to monitor the flurry of `getNameByHash` calls on boot

mslipper:
LGTM. Let me know when you want a merge.

pinheadmz:
@mslipper added one last patchset. This replaces the `state.wallet.transactions` array with a Map. The Map is still sorted by date, and gives us the feature of being able to test it for a certain txid. With that in place, another round of redundant API calls can be avoided.

Pushing this up for now, going to bang on it a bit more tomorrow and make sure it's all good.

app/ducks/walletReducer.js
```diff
@@ -56,7 +56,7 @@ export default function walletReducer(state = getInitialState(), {type, payload}
           ...state.balance
         },
         isLocked: true,
-        transactions: []
+        transactions: new Map()
```
mslipper, line 59: Curious: what is the advantage of using a `Map` in this context as opposed to a simple `{}`?

pinheadmz: The PropTypes checker was throwing an error before the first update.

pinheadmz:
rebase to 216a858cff00d5194c9109a37a89effa2f85a836

Remove leftover `console.log` from debugging and force-push

Issue #104, opened 2020-03-20
Passphrase timeout if GUI active too long


pinheadmz:
Bob requires a wallet passphrase on boot but (I think) doesn't ask for it again during wallet operations. I'm not sure how long the wallet passphrase stays in memory, but it seems like if the GUI is active long enough the passphrase is cleared and required to (e.g.) `sendupdate`.

Is there a way to pop your passphrase back in? Or we could pop open a dialog.

PR #128, opened 2020-04-21
Separate idle modal and Bob "lock" from underlying hsd wallet lock. Request user passphrase when needed.

pinheadmz:
Closes https://github.com/kyokan/bob-wallet/issues/104

### Changes

- Removes logic in `doPoll` that logs out of Bob if hsd wallet is locked
- Adds new passphrase modal that accepts callback (Promise.resolve) and can interrupt any process to unlock the hsd wallet, then return to the caller.
- Various cleanup and nitpicks to construct new modal

The new UX will be:
- Passphrase required to log in to wallet (for privacy, no change)
- Once logged in, after 5 minutes of idling, Bob logs out and locks hsd wallet (no change)
- Any time private key is needed (send, open, bid, reveal, renew, update, redeem) the modal pops up and calls `wallet.unlock()` (new)
- Any time wallet is unlocked, the default timeout (60 seconds) is applied. There is no feedback to the user when the time expires and the wallet is locked. That won't be necessary anyway because we always get the passphrase again whenever we actually need it.

### Demonstration

Demonstrating closing modal without passphrase (error is displayed to user), entering wrong passphrase, and entering correct passphrase.

require-password

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 with "spendable" balance which is total unconfirmed minus total locked coins
 - Covenants in portfolio view now display their value as it affects spendable balance
 - Improvements to maxSend based on spendable balance and cleaner fee estimation
+- Bob will ask user for passphrase whenever private key is needed (e.g. send TX)
+- Bob will no longer "logout" when underlying hsd wallet is locked, however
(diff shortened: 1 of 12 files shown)


mslipper:
LGTM, just resolve conflicts and I'll merge.

pinheadmz:
rebased on master!

mslipper:
Boom, thanks!

Issue #136, opened 2020-05-08
"failed to open bid" should say "auction"


pinheadmz:
![Screen Shot 2020-05-08 at 7 08 10 PM](https://user-images.githubusercontent.com/2084648/81456141-50b23b80-915f-11ea-8ce8-6b35877f2e60.png)

I thought I changed all these to say "open auction" instead of "open bid"

like...

![Screen Shot 2020-05-08 at 7 09 23 PM](https://user-images.githubusercontent.com/2084648/81456184-75a6ae80-915f-11ea-9aaf-56cd67b81083.png)

...and...

![Screen Shot 2020-05-08 at 7 09 56 PM](https://user-images.githubusercontent.com/2084648/81456215-87885180-915f-11ea-871b-77731480e39e.png)

Issue #135, opened 2020-05-08
Lots of names in watchlist seem stuck on "loading"


pinheadmz:
![Screen Shot 2020-05-08 at 7 03 58 PM](https://user-images.githubusercontent.com/2084648/81455954-c9fd5e80-915e-11ea-9bee-ad9623ba2362.png)

Will need to dig this one out too, not sure what happened there.

mslipper:
Maybe there were some HSD API changes?

Issue #134, opened 2020-05-08
Name registers should show amount of change returned to wallet


pinheadmz:
![Screen Shot 2020-05-08 at 7 00 37 PM](https://user-images.githubusercontent.com/2084648/81455809-47749f00-915e-11ea-8b7e-48151b269fdb.png)

This is an auction I won (with no extra lockup) where the 2nd highest bid was 50 HNS. Bob is showing the "update" as a 0-delta balance because it ignores _change_ from all transactions. In the case of a register, the difference between the winning bid and the second highest is the change output!

oops sorry, will fix.

PR #137, opened 2020-05-11
Fix bugs introduced in recent updates

pinheadmz:
Closes https://github.com/kyokan/bob-wallet/issues/134
Closes https://github.com/kyokan/bob-wallet/issues/135
Closes https://github.com/kyokan/bob-wallet/issues/136

Sorry for the monolith PR.

# 1: Reword "bid" to "auction"

# 2: Display spendable balance delta from REGISTER

### Revealing a masked bid, then registering a bid and receiving change from the second-highest bid:

Screen Shot 2020-05-11 at 10 47 47 AM

### Revealing an un-masked bid, then redeeming a losing bid:

Screen Shot 2020-05-11 at 10 46 35 AM

### Revealing an un-masked bid then registering a TIED bid for highest:

Screen Shot 2020-05-11 at 10 46 14 AM

### Revealing an un-masked bid then registering a winner with no other bids:

Screen Shot 2020-05-11 at 10 46 05 AM

# 3: Fix watchlist

https://github.com/kyokan/bob-wallet/pull/126 improved the way bids are handled but broke the way the watchlist and those name status are handled. They share a lot of the same logic, so just needed to apply the same process.

Also noticed another issue which is behavior when switching networks. I'm not sure it was working correctly before, because of the state in `doPoll()`. So now when we switch networks, everything is loaded into the store. The result is a slightly longer "loading node" time on boot, bt then everything else is nice and snappy.

### DIFF
--- a/app/components/Transactions/Transaction/index.js
+++ b/app/components/Transactions/Transaction/index.js
@@ -18,6 +18,7 @@ const UPDATE = 'UPDATE';
 const RENEW = 'RENEW';
 const REDEEM = 'REDEEM';
 const COINBASE = 'COINBASE';
+const REGISTER = 'REGISTER';
 
 class Transaction extends Component {
   static propTypes = {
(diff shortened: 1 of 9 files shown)


mslipper:
LGTM

PR #146, opened 2020-06-21
Add transfer support

mslipper:
Screen Shot 2020-06-21 at 4 55 00 PM
Screen Shot 2020-06-21 at 4 54 42 PM
Screen Shot 2020-06-21 at 4 54 02 PM
Screen Shot 2020-06-21 at 4 53 48 PM


pinheadmz:
Before digging into the code I tried out the new UX on Windows.

- First time I tried to transfer I got an error "Name must be registered." This makes sense protocol-wise because the name is still in a REVEAL, which has the wrong UTXO value (highest, not second-highest bid). So we could either disable transfer until register is done, or add a helpful message on what register means OR -- it might be possible to just send both TXs at once and the user wont even have to care. It might MIGHT even be possible to send both the REGISTER and the TRANSFER in the same TX! The transfer output would have to be after the register in the output array, but I actually think this would be valid on chain. Either way, we can save some hair-pulling FAQs here by making it easier on the user.

Also take a look at this PR: https://github.com/handshake-org/hsd/pull/464

We may have an issue with the "locked" balance after the finalize is issued (in both sender and receiver wallets...) that PR needs review still but I don't think Bob should merge transfers without this patch.

(side note: from PORTFOLIO when we click on a name in the TX list, I sorta feel like if the auction is over and won, it should navigate directly to manage domain, not the auction page...)

- In portfolio view, the FINALIZE transaction is displayed as "sent funds" without the name. I think it should say "Finalized transfer" and should use the black `0.000000` as value without + or - (wallet spendable balance is not affected)

- Also I tried transferring a name back IN to Bob, and the TX comes up in portfolio view as "Unknown Transaction" -- this would be the receipt of a FINALIZE and should say "received name". The name shows up in the domain manager just fine, etc.

pinheadmz:
Checked the UX of the REVOKE path too - looks good!

app/ducks/names.js
```diff
@@ -2,11 +2,11 @@ import nodeClient from '../utils/nodeClient';
 import walletClient from '../utils/walletClient';
 import * as namesDb from '../db/names';
 import {
+  fetchPendingTransactions,
+  getPassphrase,
```
pinheadmz, line 6: alphabetical order? 🤔 😄

app/components/Transactions/Transaction/index.js
```diff
@@ -44,19 +46,20 @@ class Transaction extends Component {
...
+          || tx.type === REDEEM
+          || tx.type === REGISTER)
+        && !tx.pending,
       'transaction__number--neutral':
         (tx.type === UPDATE
-        || tx.type === RENEW
-        || tx.type === OPEN)
+          || tx.type === RENEW
+          || tx.type === OPEN
+          || tx.type === TRANSFER)
```
pinheadmz, line 58: I think FINALIZE would go here, same as below we need a FINALIZE clause after line 117

app/ducks/namesReducer.js
```diff
@@ -9,6 +9,7 @@ const ALLOWED_COVENANTS = new Set([
   'REGISTER',
   'RENEW',
   'REDEEM',
+  'TRANSFER',
```
pinheadmz, line 12: I forget what this enum is for, but probably needs FINALIZE here also

app/ducks/walletActions.js
```diff
@@ -283,7 +282,8 @@ async function parseInputsOutputs(net, tx) {
     // affect the spendable balance of the wallet.
     // TODO: Transfer, Finalize, etc will eventually need to go here
     if (covenant.action === 'RENEW'
-        || covenant.action === 'UPDATE') {
+      || covenant.action === 'UPDATE' ||
+      covenant.action === 'TRANSFER') {
```
pinheadmz, line 286: FINALIZE here, spendable balance delta is zero.
Can remove comment on line 283!

app/ducks/walletActions.js
```diff
@@ -369,6 +369,20 @@ async function parseCovenant(net, covenant) {
...
+          domain: await nameByHash(net, covenant),
+        },
+      };
+    case 'REVOKE':
+      return {
+        type: 'REVOKE',
+        meta: {
+          domain: await nameByHash(net, covenant),
+        },
+      };
```
pinheadmz, line 385: FINALIZE here

pinheadmz:
One other thing occurred to me about this - a REVOKE is possible as soon as the TRANSFER is confirmed. Looking at youe screen shots, I think the REVOKE option only pops up after 2 days. Really, it should be a lot sooner since the point is to destroy a name *before* the attacker can FINALIZE.

sdtsui:
@mslipper after a transfer is completed (I pulled down this branch and attempted one with a throwaway name on mainnet), I am no longer able to update records -- it says I need to register first.


Screen Shot 2020-06-27 at 2 31 36 AM

If this is a bug and I don't actually need to register, let's fix it before merging?
If it's not a bug, and I do need to register, is it possible to confirm receipt somehow without scanning the whole chain?  Maybe a button that allows the register to be sent from the new address.


also, in the process of testing this, I noticed some copy inaccuracies: they are in #150

app/ducks/namesReducer.js
mslipper: It's used to keep track of which covenants the wallet registers pending transactions for.

PR #148, opened 2020-06-25
Allow importing 128-bit entropy seeds

HarukaMa:
Wallets created using old hsd nodes have 128-bit entropy, thus the seed phrase only have 12 words.

This PR allows importing such wallets.

### DIFF
--- a/app/pages/Onboarding/ImportSeedEnterMnemonic/index.js
+++ b/app/pages/Onboarding/ImportSeedEnterMnemonic/index.js
@@ -75,7 +75,7 @@ export default class ImportSeedEnterMnemonic extends Component {
   }
 
   disableButton() {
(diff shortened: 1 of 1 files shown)


sdtsui:
@mslipper can you please give this a QA run?

The first time I ran this branch, it silently failed after clicking "Continue" in the analytics opt-in screen (Step 4).  "Continue" was unselected again. There was an error output but I cleared my terminal before I could catch it.

After that, every subsequent time I ran `npm run dev` I got the following state for Bob.

Screen Shot 2020-07-02 at 6 00 19 PM

Of course, these might be vanilla build issues but I don't yet have enough context w/ the codebase to merge even though the change looks decent.

sdtsui:
Also: thank you HarukaMa.

mslipper:
I was not able to reproduce the issue described above. Merging.

Issue #163, opened 2020-07-16
Issues with bids of 0 value


pinheadmz:
h/t @falci

Bob doesn't normally allow bids of `0` value, but this can be sidestepped by entering a value of `0.00000000001`:

https://github.com/kyokan/bob-wallet/blob/4c424a19580ababf08ac7b5bd22b92aebdd4c844/app/pages/Auction/BidActionPanel/BidNow.js#L189

This is one thing that should be fixed -- 6 decimal place max, as enforced here:

https://github.com/kyokan/bob-wallet/blob/65ae2106cf82624b240a2efd89cd269c85567074/app/pages/Auction/RepairBid.js#L58

The user can also use RPC calls to send a bid of `0`, which is totally valid **as long as the lockup is above the dust limit**

Unfortunately the `0` value is incorrectly interpreted as false in two places which both trigger repair bid, and prevent repair bid:

Incorrectly flagging a bid of `0` as missing (in need of repair -- this should just check for `undefined`):

https://github.com/kyokan/bob-wallet/blob/4c424a19580ababf08ac7b5bd22b92aebdd4c844/app/pages/Auction/BidActionPanel/BidNow.js#L411

Preventing value of `0` being entered as a repaired bid (this should just be a type check):

https://github.com/kyokan/bob-wallet/blob/65ae2106cf82624b240a2efd89cd269c85567074/app/pages/Auction/RepairBid.js#L62

Falci:
My 2 cents about the bid value:
* If the network doesn't store the information, not even encoded, should Bob display it? (it seems a "privileged information")
* And the opposite: if it's an interesting information, shouldn't the network store it? Maybe encoded?

pinheadmz:
The bids are not stored on the network because they are blind! That is the design of the system. Bob only displays YOUR bids because they are your secret, and they are also required for the REVEAL, when the bid value is actually plainly broadcast to the network. I have some ideas on how to make the bid value recoverable like the rest of the BIP44 scheme: https://github.com/handshake-org/hsd/issues/378#issuecomment-618486570

brandondees:
Since this is electron we could use html5 input validation to make it even less accident prone, right? (I mean reference the same regex from the input, so nobody ever sees an error without using hacker methods, they can't even push the button till it's valid).

pinheadmz:
@brandondees I'm not sure exactly what you mean, but I have written a patch for this already if you want to take a look:

https://github.com/kyokan/bob-wallet/pull/170

[...]

PR #165, opened 2020-07-17
Allow 0 as "Repair bid" value

Falci:
Related to #163 

This changes depends on handshake-org/hsd/pull/477. 
Without the update on hsd, typing "0.000001" will show an error on each 0, removing the field focus (terrible UX);


sdtsui:
@pinheadmz do you mind giving this a review?  Shoot me a DM if you want write access so that hsd/477 can be merged concurrently.

I suppose we'll also want a version bump in here?  If we do it this way, we'll want to make sure the lookahead var in our fork stays higher as some of our users have a lot of stored names.

app/pages/Auction/RepairBid.js
```diff
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import {consensus} from 'hsd/lib/protocol';
 import walletClient from '../../utils/walletClient';
 import * as names from '../../ducks/names';
+import { showError } from '../../ducks/notifications';
```
pinheadmz, line 7: I don't think this is necessary, `showError` is connected a few lines down

app/pages/Auction/RepairBid.js
```diff
@@ -50,17 +51,19 @@ export default class RepairBid extends Component {
...
     const value = val.match(/[0-9]*\.?[0-9]{0,6}/g)[0];
-    if (value * consensus.COIN > consensus.MAX_MONEY)
+    const dollarydoos = parseFloat(value) * consensus.COIN;
+    if (dollarydoos > consensus.MAX_MONEY)
       return;
     this.setState({value: value});
-    if (parseFloat(value))
-      return this.verifyBid(value);
+    if (val === "")
+      return;
```
pinheadmz, line 65: Mine as well move this to the top of the function I think.

pinheadmz:
@Falci do you have the time to finish the other bugs in #163? If not I can add commits to this PR. I think we should nail it all in one move

app/pages/Auction/RepairBid.js
```diff
@@ -50,17 +51,19 @@ export default class RepairBid extends Component {
...
         disabled={this.state.isCorrect}
-      />      
+      />
     );
   }
 
   processValue = async (val) => {
     const value = val.match(/[0-9]*\.?[0-9]{0,6}/g)[0];
-    if (value * consensus.COIN > consensus.MAX_MONEY)
+    const dollarydoos = parseFloat(value) * consensus.COIN;
```
pinheadmz, line 60: We still need a type check here. If the user enters for example `"foobar"` then `parseFloat()` returns `NaN` which will throw an error during `setState()`. Suggestion, maybe we need this:

```js
const dollarydoos = parseFloat(value);
if (!Number.isNaN(dollarydoos))
  return;
```

app/pages/Auction/RepairBid.js
Falci: It seems like you forgot to import if before connecting. 
Without it, there is an console error, about `showError` not being defined.

app/pages/Auction/RepairBid.js
Falci: `foobar` wouldn't pass the regex.

app/pages/Auction/RepairBid.js
Falci: This needs to be after the `setState`. Otherwise, user can't clear the input.

pinheadmz: great, thanks

app/pages/Auction/RepairBid.js
pinheadmz: ok cool thank you

Falci:
> @Falci do you have the time to finish the other bugs in #163? If not I can add commits to this PR. I think we should nail it all in one move

Do you mean enable 0 bid? I'm afraid I won't have time.

app/pages/Auction/RepairBid.js
```diff
@@ -69,7 +72,7 @@ export default class RepairBid extends Component {
       const attempt = await walletClient.getNonce({
         name: bid.name,
         address: bid.from,
-        bid: value * consensus.COIN
+        bid: value
```
pinheadmz, line 75: Ok one more thing: we shouldn't change the value sent to `verifyBid()` because a few lines down `rpc importNonce` actually requires a float in HNS, not an integer in dollarydoos. The HTTP endpoint `getNonce` DOES accept an integer of dollarydoos so your intuition is correct, but this will break repairing any bid that is NOT `0`.

For this reason, I think we need to keep this line as `bid: value * consensus.COIN` and up on line 66, weneed to keep `return this.verifyBid(value);`

app/pages/Auction/RepairBid.js
```diff
@@ -57,13 +57,13 @@ export default class RepairBid extends Component {
...
   processValue = async (val) => {
     const value = val.match(/[0-9]*\.?[0-9]{0,6}/g)[0];
-    const dollarydoos = parseFloat(value) * consensus.COIN;
-    if (dollarydoos > consensus.MAX_MONEY)
-      return;
     this.setState({value: value});
-    if (val === "")
+    const parsed = parseFloat(value);
+
+    if (val === "" || Number.isNaN(parsed) || parsed * consensus.COIN > consensus.MAX_MONEY)
```
pinheadmz, line 63: I think `setState` needs to come after this check, so if the user tries to type `999999999999999`, it will just freeze once they get to MAX_MONEY, etc.

pinheadmz:
https://github.com/handshake-org/hsd/pull/477 is merged, so this PR is now unblocked.

Optionally could be released with https://github.com/kyokan/bob-wallet/pull/170 which is blocked by https://github.com/handshake-org/hsd/pull/479

app/pages/Auction/RepairBid.js
Falci: Nothing happens if you type a value too high.
The user doesn't know that we try to validate on each key pressed. In this case, we just don't try to validate.
It's harmless and transparent.

pinheadmz: ok fair enough!

Falci:
Do I need to do anything else to merge this?

pinheadmz:
> Do I need to do anything else to merge this?

Just need to wait for repo maintainer review and approval :-) @mslipper @sdtsui @chikeichan

chikeichan:
This PR looks good to me -- thanks @Falci !

@pinheadmz since Bob uses a forked version of `hsd` that uses a larger lookahead value, would this require us to update and cut a new version of `hsd`?

chikeichan:
Used this branch for a couple of days and were able to verify the following:
- able to submit bid with local (http) custom RPC
- able to reveal bid with normal p2p
- able to send tx with remote (https) custom RPC

@pinheadmz just want to see if you were able to run this branch, and if you are okay with merging this 😃

pinheadmz:
@chikeichan https://github.com/handshake-org/hsd/pull/477 is required for this patch to operate correctly in Bob. You could either cherry-pick that patch into the Kyokan hsd fork, or rebase your own patch set on hsd master before each Bob release.

To be honest, I think a PR to bump hsd lookahead to 1000 would probably be supported by the developers and get merged, even though it's a dumb "kick the can down the road" fix. But everyone with busy wallets is making this modification on their own anyway...

Related, allow Bob users to directly place 0-value bids: https://github.com/kyokan/bob-wallet/pull/170
This would require https://github.com/handshake-org/hsd/pull/460/files (already merged to hsd) and https://github.com/handshake-org/hsd/pull/479 (in review) to protect users from footgunning a dust output.

Probably https://github.com/handshake-org/hsd/pull/479 should be merged before next Bob release as well.

chikeichan:
@pinheadmz thanks for the context! There is actually a [setLookahead](https://github.com/handshake-org/hsd/blob/master/lib/wallet/account.js#L566) method in the account object. We just need to query the object directly from `wdb` instead of getting it from wallet rpc, which is made a bit easier now that Bob is not using the wallet as a plugin. I just tested it and it allows one of my wallet with large lookahead to sync. Do you see potential problem with using this method?

It also seems like `hsd` has bumped `bcrypto` to match the [version](https://github.com/bcoin-org/bcrypto/32e6ce4c189e9c748373ce89bc6c7949f9a2992b) in our forked.

We likely don't need to maintain a fork anymore. I will file a issue to remind myself to verify and remove the fork when we get closer to next release.

pinheadmz:
> @pinheadmz thanks for the context! There is actually a [setLookahead](https://github.com/handshake-org/hsd/blob/master/lib/wallet/account.js#L566) method in the account object. We just need to query the object directly from `wdb` instead of getting it from wallet rpc, which is made a bit easier now that Bob is not using the wallet as a plugin. I just tested it and it allows one of my wallet with large lookahead to sync. Do you see potential problem with using this method?

hm interesting - the `if (lookahead < this.lookahead)` condition looks a little funny to me in this block - like if the original value is `10` and you pass in `1` -- it will just add `9` to the `depth` values and return? 🤔 

So you're saying in Bob you just call this method "manually" and bump the lookahead without having to change the code in hsd? That actually looks ok -- and in fact you may only need to call that once after you create the account (the `"default"` account is created when you first create a wallet). The `lookahead` value is serialized and saved with the rest of the account metadata. 

The only issue I see with that is if someone uses unpatched hsd to restore the wallet. The default lookahead will be `10` again and they may not get a full BIP44 recovery. I think your method is ok for now but we should still look at bumping this value for all hsd.

chikeichan:
yea it seems like the lookahead values also affects depths, so there is some special handling of the change in lookahead value to correctly adjust depth with the same change. i also verified you can only increase lookahead value, but not decrease it.

> The only issue I see with that is if someone uses unpatched hsd to restore the wallet. The default lookahead will be 10 again and they may not get a full BIP44 recovery. I think your method is ok for now but we should still look at bumping this value for all hsd.

almost seems like `hsd` would want this to be a config passed in thru CLI and the constructor as an option, or perhaps exposing this method thru RPC? (actually, exposing this method thru RPC seems pretty straightforward)

pinheadmz:
The reason it's not an option or an RPC is because it is supposed to be a standard. [BIP44 actually defines the standard lookahead as 20](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit). We don't really want users to have to remember an extra integer to correctly restore their wallet. Let's continue this here: https://github.com/handshake-org/hsd/issues/407

Falci:
Hi,
There are 2 issues in #163: repair 0 bids and allow 0 bid.
Keep in mind this PR is just about the repair part.

chikeichan (approved):
this should fix #221

PR #167, opened 2020-07-18
Add button to export names

sdtsui:
This feature has been requested by several power users.  We would like to add batch functionality later in the year, ideally with .csv as an input source.  For now, this button allows users to export all of their names in Bob for sale on secondary markets, piping into scripts, etc.

- New Button:
Screen Shot 2020-07-17 at 10 53 15 PM

- When clicked:
Screen Shot 2020-07-17 at 10 53 29 PM

- Example output:
Screen Shot 2020-07-17 at 11 05 55 PM

@mslipper @chikeichan - it's been a while since I've contributed directly to Bob.  Please let me know if I'm breaking any previously established patterns or conventions; links to specific folders or templates to emulate would be great if so.  I'll be working on smaller features until I get a better feel for the codebase.


app/pages/DomainManager/index.js
```diff
@@ -22,9 +25,32 @@ class DomainManager extends Component {
...
+    let data = names.join("\n");
+
+    let savePath = dialog.showSaveDialogSync({
+      filters: [{ name: 'spreadsheet', extensions: ['csv'] }]
+    });
+
+    if(savePath) {
+      fs.writeFile(savePath, data, (err) => {
+        if (err) {
+          throw err;
```
chikeichan, line 39: can you add error handling here? you can refer to the example here: https://github.com/kyokan/bob-wallet/blob/master/app/pages/Auction/BidActionPanel/Reveal.js#L61-L63

you will also have to add a new md2p: https://github.com/kyokan/bob-wallet/blob/master/app/pages/Auction/BidActionPanel/Reveal.js#L193

app/pages/DomainManager/index.js
```diff
@@ -22,9 +25,32 @@ class DomainManager extends Component {
...
+  handleExportClick() {
+    let names = this.props.myDomains.map(({name}) => name);
+    let data = names.join("\n");
+
+    let savePath = dialog.showSaveDialogSync({
+      filters: [{ name: 'spreadsheet', extensions: ['csv'] }]
+    });
+
+    if(savePath) {
+      fs.writeFile(savePath, data, (err) => {
```
chikeichan, line 37: you can actually create a csv in memory and trigger a download. Check out the method use here: https://github.com/kyokan/bob-wallet/blob/master/app/pages/Watching/index.js#L33-L41

app/pages/DomainManager/index.js
sdtsui: @chikeichan - yep the sendReveal example makes sense

what is an md2p? are we making schedule I substances?

app/pages/DomainManager/index.js
sdtsui: hmm, would you rather i use the watchlist code?

also, wdyt of centering the "download watchlist" button instead of the big blue extension button - which do you think looks better?

chikeichan: yea let's use the watchlist code because it's the same functionality -- even if you just do copy-and-paste instead of refactoring into a utility function, it would be easier to maintain.

what do you mean by "_centering download watchlist button_"?

PR #171, opened 2020-07-22
Add custom rpc and new settings view

chikeichan:
This PR adds the ability to stop the local node and point RPC to another host.

I manually tested the following flows:
- upgrade from previous version of Bob doesn't require app directory wipe
- network type persists when app reloads
- re-import wallet with custom rpc
- finalize a transfer with custom rpc

related to custom rpc feature, I plan on adding to the next PR:
- ability to connect to any hosted endpoint (https)

@sdtsui would love a run through of this and any suggestions on UI improvement.

networkswitch


pinheadmz:
It is CRITICAL that users only connect Bob to a local hsd node, either same computer or same LAN behind a firewall. If users attempt to run an hsd node on a VPS, with `http-host=0.0.0.0` to be controlled with Bob they ABSOLUTELY MUST set up SSL on the server side. This is a slightly advanced hsd [configuration ](https://hsd-dev.org/guides/config.html):

```
ssl: true
ssl-cert: /path/to/cert.crt
ssl-key: /path/to/privatekey.key
```

And then in Bob, they can not use an IP address for host, but an https URL.

If the hsd server does not have a legacy domain name with legacy certificate, a self-signed certificate may be used but then it also must be installed on the OS that is running Bob (although there are cool nodejs tricks to add an extra cert at runtime without asking the OS for permission: note the `ca` option in https://nodejs.org/api/https.html#https_https_request_options_callback)

Failure to enforce SSL on a remote node means the API key is transmitted in plaintext, the user is giving away their money to everyone on the internet!

This information can either be added in a big red warning, or Bob can be configured to only allow `127.0.0.0/32` and `10.0.0.0/32` type addresses

pinheadmz:
Regarding use of `WalletNode` - this is actually suboptimal if the full node and wallet are on the same machine, and will cause additional issues with rescanning and importing from seed. The reason is that as a separate process, wallet and node communicate over HTTP API and websockets instead of just direct javascript function calls and shared objects. In particular, a remote wallet relies on a bloom filter like an SPV wallet to get transactions from the full node. The bloom filter does not update frequently enough during rescan / import from seed: 

https://github.com/bcoin-org/bcoin/issues/835

https://github.com/bcoin-org/bcoin/issues/842

For this reason I recommend leaving the default configuration as-is (where hsd starts with wallet as a plugin) if the user wants to switch to a remote wallet, hsd can be restarted with `--no-wallet` or `no-wallet: true` .

pinheadmz:
Does this also assume the remote hsd+wallet server has a wallet named `allison`? What if it doesn't? I can imagine a user running hsd with the default `primary` wallet for a while then deciding they want to run Bob as a remote control - maybe the settings screen should allow a custom wallet name as well?

pinheadmz:
When hsd is running already, Bob starts up with the "ports in use error" -- I wonder if this error message can also have a link to "switch to your own node"?

pinheadmz:
I'm having trouble running this branch actually: Can not get past the "already have a wallet"->"import seed" screen (stuck on "creating wallet..." or the "create a new wallet"->"I agree" screen (stuck on "loading...")

This is with a fresh `npm i` and empty data directory.

Electron output:

```

Sending IPC method response. {
  id: 41,
  jsonrpc: '2.0',
  method: 'DB.get',
  params: [ 'initialization-state:regtest' ]
} null
Executing IPC method. {
  id: 42,
  jsonrpc: '2.0',
  method: 'DB.get',
  params: [ 'initialization-state:regtest' ]
}
```

Regtest hsd debug log looks ok, I see periodic rpc calls to `estimatefee`. In mainnet mode, hsd seems to be syncing the chain fine, but I still can not advance past the wallet creation / restore screen.

brandondees:
subscribing to track progress, I'm just eager to see a version of this released

chikeichan:
> I think it is important that users only connect Bob to a local hsd node, either same computer or same LAN behind a firewall. If users attempt to run an hsd node on a VPS, with `http-host=0.0.0.0` and `wallet-http-host=0.0.0.0` to be controlled with Bob they really should set up SSL on the server side. This is a slightly advanced hsd [configuration ](https://hsd-dev.org/guides/config.html):
> 
> ```
> ssl: true
> ssl-cert: /path/to/cert.crt
> ssl-key: /path/to/privatekey.key
> ```
> 
> And then in Bob, they can not use an IP address for host, but an https URL.
> 
> If the hsd server does not have a legacy domain name with legacy certificate, a self-signed certificate may be used but then it also must be installed on the OS that is running Bob (although there are cool nodejs tricks to add an extra cert at runtime without asking the OS for permission: note the `ca` option in https://nodejs.org/api/https.html#https_https_request_options_callback)
> 
> Failure to enforce SSL on a remote node means the API key is transmitted in plaintext, the user is giving away their money to everyone on the internet!
> 
> This information can either be added in a big red warning, or Bob can be configured to only allow `127.0.0.0/32` and `10.0.0.0/32` type addresses

Nice suggestion -- I changed the custom rpc ux to show a warning if user try to connect to remote address without https

warning

chikeichan:
> Does this also assume the remote hsd+wallet server has a wallet named allison?

it assume the wallet server (running in bob) has a wallet named `allison`. I have also checked the rpc, http, and socket calls in the `hsd` node. It doesn't seem to need to know the correct wallet id to work, but I could be missing stuffs 😅 . Do you know if anything would break in this case?

chikeichan:
> When hsd is running already, Bob starts up with the "ports in use error" -- I wonder if this error message can also have a link to "switch to your own node"? Same with setting up a node for the first time -- it goes through the entire 24-word seed flow before you have the option to switch to your own node.

Yea this sounds right -- I will include #182  in my next couple PRs

chikeichan:
> Regarding use of `WalletNode` - this may be suboptimal if the full node and wallet are on the same machine, and could cause additional issues with rescanning and importing from seed. The reason is that as a separate process, wallet and node communicate over HTTP API and websockets instead of just direct javascript function calls and shared objects. In particular, a remote wallet relies on a bloom filter like an SPV wallet to get transactions from the full node. The bloom filter does not update frequently enough during rescan / import from seed:
> 
> [bcoin-org/bcoin#835](https://github.com/bcoin-org/bcoin/issues/835)
> 
> [bcoin-org/bcoin#842](https://github.com/bcoin-org/bcoin/issues/842)
> 
> For this reason I think leaving the default configuration as-is (where hsd starts with wallet as a plugin) if the user wants to switch to a remote wallet while using the Bob internal full node, hsd can be restarted with `--no-wallet` or `no-wallet: true` .
> 
> Or just disable here:
> 
> https://github.com/kyokan/bob-wallet/blob/4f4cc73ea7075fe04c1fd128e9bef6797fd86469/app/hsdMain.js#L35

Took me a few days but it should be working now! The problem did mostly come from sockets and making sure the wallet server and the node is communicating properly. Give you a try again if you have time 😃

pinheadmz:
Cool, thanks for the responses. I admit I was confused about the design and that's why I was confused about `WalletNode` and `"allison"` -- I thought the goal was to configure Bob as a totally empty UI for a remote hsd that had its own wallet plugin already (in which case, the wallet would probably be named `"primary"`). 

So I understand now that the goal is to host a wallet node only inside Bob and allow switching from the internal hsd full node to a remote hsd full node. There still may be issues with the `WalletNode` configuration for ordinary users running the internal full node (see those two bcoin issues) especially during rescan or import seed. Especially on Windows which still currently uses the slow JS crypto backend, the rescan calls may timeout.

I'm also not entirely sure how well a remote hsd full node (lets say one hosted by Kyokan) will handle multiple wallet clients. I'm afraid if one client initiates a rescan, it may lock the full node for the duration, for everybody. For this reason, it may even make more sense to just configure Bob as an hsd SPV node with wallet plugin, or build some other type of Electrum-ish API if Kyokan plans on serving multiple wallet clients at once.

Regarding wallet names, yeah -- all the HTTP API calls take an `--id=...` parameter with `"primary"` as the default. That is also the default for the wallet RPC calls, but the wallet RPC is stateful and so `rpc selectwallet allison` can be set once and then all future API calls will follow. In fact if you have Bob running and the API key set in a local terminal, you can call `hsw-rpc selectwallet primary` and Bob will freak out because all the `"allison"` data is gone and all the `getHistory` calls, etc will return empty. So now I realize that you're keeping the wallet local anyway and you won't have this issue. But if you ever configure Bob as an empty UI to control a complete remote hsd+wallet plugin, the wallet name will be necessary.

And actually, I know some users have requested multi-wallet support (say to switch between their airdrop seed and the new seed Bob generates) and so enabling switching wallets by name will eventually be a cool thing to have anyway. Wallet switching also means users will be wiping/reimporting from seed a lot less, and that process causes a lot of headaches!

pinheadmz:
Had to 

```diff

**Comment by pinheadmz** [comment]
Still getting this trying to `npm i`:

```
npm ERR! code ENOLOCAL
npm ERR! Could not install from "node_modules/hsd/bcrypto@git+https:/github.com/bcoin-org/bcrypto.git#5c3e2b8b3f79d13e1b5c26be197708d0b9e5bcf9" as it does not contain a package.json file.
```

fixed:

```diff
diff --git a/package.json b/package.json
index bb0c814..2f2388a 100644
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "electron-updater": "4.1.2",
     "history": "4.7.2",
     "hs-client": "https://github.com/handshake-org/hs-client.git",
-    "hsd": "https://github.com/kyokan/hsd/tarball/296f36c996d9e1561fd0718d9887402181d57c85",
+    "hsd": "https://github.com/kyokan/hsd",
     "isomorphic-fetch": "^2.2.1",
     "lodash.isequal": "4.5.0",
     "lodash.throttle": "4.1.1",
```

pinheadmz:
Hm, can't get past "Loading node..." now: (`npm run dev`)

```
Starting node on main network.
Executing IPC method. {
  id: 7,
  jsonrpc: '2.0',
  method: 'DB.get',
  params: [ 'initialization-state:main' ]
}
Sending IPC method response. {
  id: 7,
  jsonrpc: '2.0',
  method: 'DB.get',
  params: [ 'initialization-state:main' ]
} null
Error: Timed out waiting for connection. Reconnecting...
    at Socket.error (/Users/matthewzipkin/Desktop/work/bob-wallet/node_modules/bsock/lib/socket.js:327:24)
    at Socket.stall (/Users/matthewzipkin/Desktop/work/bob-wallet/node_modules/bsock/lib/socket.js:393:14)
    at Timeout._onTimeout (/Users/matthewzipkin/Desktop/work/bob-wallet/node_modules/bsock/lib/socket.js:370:41)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7)
```

chikeichan:
> I'm also not entirely sure how well a remote hsd full node (lets say one hosted by Kyokan) will handle multiple wallet clients. I'm afraid if one client initiates a rescan, it may lock the full node for the duration, for everybody. For this reason, it may even make more sense to just configure Bob as an hsd SPV node with wallet plugin, or build some other type of Electrum-ish API if Kyokan plans on serving multiple wallet clients at once.

Yea figuring out how to scale to multiple wallet clients took some times, but at the end I got it working by using my own bsocket server to intercept the socket requests so that I can respond with data from redis instead of locking up the node. For configuring Bob as SPV node, i think that's the end goal (or at least make it an advanced option) in the future as the chain gets longer.

chikeichan:
> Still getting this trying to `npm i`:
> 
> ```
> npm ERR! code ENOLOCAL
> npm ERR! Could not install from "node_modules/hsd/bcrypto@git+https:/github.com/bcoin-org/bcrypto.git#5c3e2b8b3f79d13e1b5c26be197708d0b9e5bcf9" as it does not contain a package.json file.
> ```
> 
> fixed:
> 
> ```diff
> diff --git a/package.json b/package.json
> index bb0c814..2f2388a 100644
> --- a/package.json
> +++ b/package.json
> @@ -119,7 +119,7 @@
>      "electron-updater": "4.1.2",
>      "history": "4.7.2",
>      "hs-client": "https://github.com/handshake-org/hs-client.git",
> -    "hsd": "https://github.com/kyokan/hsd/tarball/296f36c996d9e1561fd0718d9887402181d57c85",
> +    "hsd": "https://github.com/kyokan/hsd",
>      "isomorphic-fetch": "^2.2.1",
>      "lodash.isequal": "4.5.0",
>      "lodash.throttle": "4.1.1",
> ```

Current `master` also use `"https://github.com/kyokan/hsd/tarball/296f36c996d9e1561fd0718d9887402181d57c85"` -- perhaps it's unrelated?

chikeichan:
> Hm, can't get past "Loading node..." now:

I was able to reproduce and that was actually because I accidentally removed an ipc message that tells Bob the node has been started 😅

app/pages/Settings/CustomRPCConfigModal.js
```diff
@@ -0,0 +1,148 @@
...
+        title="Configure Custom RPC"
+        centered
+      >
+        <div className="settings__input-row">
+          <div className="settings__input-title">URL</div>
+          <input
+            type="text"
+            className="settings__input"
+            value={url}
+            placeholder="http://127.0.0.1:12037"
```
pinheadmz, line 109: port number in the placeholder could be based on network setting (for example, regtest is `14037`)

chikeichan: updated!

chikeichan:
Used this branch for a couple of days and were able to verify the following:

able to submit bid with local (http) custom RPC
able to reveal bid with normal p2p
able to send tx with remote (https) custom RPC
@pinheadmz just want to see if you were able to run this branch, and if you are okay with merging this

app/background/node/service.js
```diff
@@ -227,7 +270,7 @@ async function checkHSDPortsFree(network) {
   const ports = [
     network.port,
     network.rpcPort,
-    network.walletPort,
+    // network.walletPort,
```
sdtsui, line 273: del?

sdtsui:
@chikeichan is it ok to merge this into a separate branch and cut an `experimental` release?  I'm thinking we can merge it back into master when hosted functionality is completed.

pinheadmz:
Fresh install, empty datadir, using `npm run dev`, loads OK.

 I love the new settings page layout, very clean and easy to navigate 👍 

But still having some issues.

Started fine, switched to regtest. Went to settings and clicked 'stop' -- bob logged me out and returned to "enter your seed phrase" screen again. I did so, and then I got stuck on "syncing wallet 0%" -- I didn't have a chance to enter the IP of my remote hsd node yet.

So quit and relaunched. This time I was able to go to settings and enter remote hsd info. I tried setting remote node url to a raspberry pi running hsd on my LAN by its IP address. I think there should be a timeout here because it just tried connecting forever and I must've entered somthing wrong. Couldn't cancel and try again.  I noticed this error in the log: ` Invalid URL: 10.0.1.5`

Quit and relaunch. This time stuck on "loading node...". Brick wall, every restart.

Cleaned out datadir and start over, try again. This time I set the remote node URL specifically to `http://10.0.1.5:12037` (because I also noticed in the log it was trying to hit port `80`, not the default rpc port, so without specifying port it didn't connect)

Now we're in business! I watched the remote hsd log, Bob seemed to rescanning OK - although I wasnt seeing the "syncing %" modal from https://github.com/kyokan/bob-wallet/pull/117  While rescanning, the Bob UI doesn't really work, it won't load auction details etc until the rescan job is done.

Before rescan was done though I tried something else, switching to another node. This time, local regtest. So set the URL to `http://127.0.0.1:14037` Bob didn't complain but it should have - Bob was configured for mainnet but that node is on regtest. I noticed the "network" setting is disabled when you configure a remote hsd -- I don't think it should be -- a mainnet wallet trying to rescan from a regtest node will be weirdo, and there should maybe be a check in place for that.

I'm happy to help with code review but the PR is really big (`+13,009 −160` !) It might be easier to get review if it were split up into a few smaller patches. Like
1- refactor settings page
2- use walletnode internally
3- option to switch to remote node
Just my two cents. Especially up to just try out (2) with everything from rescans to seed imports to auctions and make sure nothing has changed in Bob before adding the real feature.

chikeichan:
Thanks for the detailed feedback @pinheadmz ! I filed #188 to improve validation for the custom rpc form in the future. And sorry for the large PR -- I will stick to better practice and submit smaller PRs in the future, for this one since it's a major refactoring, let's merge in and test in a prod build.

@sdtsui -- let's merge in to `rc-0.5.0` and use that as our release build. We can then merge back to `master` after it's QA and published. Should be good enough as our release cadence for now. I will also #165 and #170 in `rc-0.5.0`

chikeichan:
Actually, #165 and #170 is blocked by https://github.com/handshake-org/hsd/pull/479

Let's include those those in next release instead. Only custom rpc for `v0.5.0`.

pinheadmz:
> Actually, #165 and #170 is blocked by [handshake-org/hsd#479](https://github.com/handshake-org/hsd/pull/479)
> 


handshake-org/hsd#479 just needs review and ACK and I can merge it! Rally some troops! 😄

app/background/wallet/service.js
```diff
@@ -269,42 +328,87 @@ class WalletService {
...
+
     this.networkName = networkName;
+    this.apiKey = apiKey;
+    this.network = network;
+
     const walletOptions = {
       network: network,
-      port: network.walletPort,
-      apiKey,
+      port: 12137,
```
pinheadmz, line 352: because this is not the standard wallet port, hs-client will not be able to communicate with Bob's wallet without an extra parameter `--http-port=12137`

app/background/wallet/service.js
```diff
@@ -269,42 +328,87 @@ class WalletService {
...
+      apiKey: crypto.randomBytes(20).toString('hex'),
```
pinheadmz, line 353: This api key is not displayed to the user anywhere? In the settings tab, the API KEY that can be revealed is probably for the full node, not the wallet (which I guess now has a different key?) This again makes using CLI tools with Bob very difficult

PR #194, opened 2020-08-06
Displays punycode next to domain name

Falci:
image

image

image

image

### DIFF
--- a/app/utils/nameHelpers.js
+++ b/app/utils/nameHelpers.js
@@ -1,3 +1,5 @@
+import punycode from 'punycode';
+
 export const states = {
   OPENING: 'OPENING',
   BIDDING: 'BIDDING',
@@ -73,4 +75,18 @@ function checkState(name, expectedState) {
   }
 
   return info.state === expectedState;
+}
+
+export const formatName = name => {
+  if (!name)
+    return name;
+
+  try {
+    const unicode = punycode.toUnicode(name);
+    if (unicode !== name) {
+      return `${name}/ (${unicode})`;
+    }
+  } catch(e) {}
+
+  return `${name}/`;
 }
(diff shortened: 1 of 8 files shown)


app/utils/nameHelpers.js
```diff
@@ -74,3 +76,17 @@ function checkState(name, expectedState) {
...
+export const formatName = name => {
+  if (!name)
+    return name;
+
+  try {
+    const unicode = punycode.toUnicode(name);
+    if (unicode !== name) {
+      return `${name}/ (${unicode})`;
+    }
+  } catch(e) {}
```
brandondees, line 89: I think the catch should probably write exceptions to the debug log, and maybe add something to the telemetry feedback to kyokan.

brandondees:
Code looks good to me overall. I may have a chance to try testing this but haven't yet. I recommend anybody doing testing check specifically for some invalid puny strings like `xn--0` which broke namebase with an uncaught exception for a while. Also try to check what happens when strings get very long, use invisible chars, etc.

app/utils/nameHelpers.js
Falci: Some names are not punycode and they can break the `.toUnicode` method. No big deal, they are not really punycode. This is why there's no track of this error.

Falci:
> xn--0

image

> what happens when strings get very long
image
image

This is the worst result I could get:
image


image
image

app/utils/nameHelpers.js
brandondees: i thought it might be useful for identifying any edge cases where the library doesn't recognize newer unicode, or things like that.

brandondees:
thanks for the screenshots, @Falci this looks great to me

brandondees:
layout/wrapping behavior for super long names is a separate issue to tackle independently, i just wanted to see whether this would cause any new forms of issues related to that (not really, it seems the emojis wrap better than ascii)

chikeichan:
Thanks @brandondees for the reviews! Merging this into rc for next release

Issue #196, opened 2020-08-06
Allow search by emoji


chikeichan:
We can allow search/filter by emoji using the `punycode` library already pulled in by #194

Falci:
It's done in the 0.5.0-rc. Can we close this issue?

PR #197, opened 2020-08-06
Implements search by punycode

Falci:
#196 

image

image


Falci:
Smells like a conflict coming...

PR #199, opened 2020-08-13
Show correct API for wallet node

chikeichan:
Hotfix for _invalid wallet api key_ issue reported by @pinheadmz on Telegram.

This add a new panel in setting for user to grab the valid api key. If the user use custom RPC, it would generate its own api key instead.

Also added a description below to remind user to always choose the wallet **allison**

Screen Shot 2020-08-12 at 8 12 28 PM

### DIFF
--- a/app/pages/Settings/index.js
+++ b/app/pages/Settings/index.js
@@ -21,6 +21,7 @@ import copy from "copy-to-clipboard";
 import {setCustomRPCStatus} from "../../ducks/node";
 import CustomRPCConfigModal from "./CustomRPCConfigModal";
 import {fetchWalletAPIKey} from "../../ducks/walletActions";
+import Anchor from "../../components/Anchor";
 
 const analytics = aClientStub(() => require('electron').ipcRenderer);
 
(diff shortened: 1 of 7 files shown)


app/pages/Settings/index.js
```diff
@@ -245,6 +250,12 @@ export default class Settings extends Component {
                 'Remove Wallet ',
                 () => history.push('/settings/wallet/new-wallet'),
               )}
+              {this.renderSection(
+                'API Key',
+                'API key for hsw-cli and hsw-rpc. Please make sure you select the wallet "allison"',
```
sdtsui, line 255: Can we find a way to link [this](https://hsd-dev.org/api-docs/#selectwallet)?   Otherwise unlikely to really inform anyone.

The users we're trying to protect are unlikely to know what it means to select a wallet, so official API docs may help them get up to speed.  Other users won't know what to make of "Allison" - sounds like it could be a setting in Bob.

pinheadmz:
Don't forget the port too: https://github.com/kyokan/bob-wallet/pull/171/files#r468608128

app/pages/Settings/index.js
pinheadmz: The api-docs intro section has a bit better explanation under the "Wallet: rpc selectwallet vs --id=..." section up top.

I think a CLI guide in general might be helpful, maybe even as a complete built in page in Bob UI. We talked about this before too, about having a built-in console entirely inside Bob, like electrum:

https://bitzuma.com/posts/an-introduction-to-the-electrum-python-console/

chikeichan: Just want to say ditto for Electrum style console for CLI and logs inside Bob. It would be a really nice addition and provide a clean way to directly use the node as fallback. In terms of UI it can simply be a tab in Settings.

chikeichan:
Screen Shot 2020-08-18 at 2 54 04 PM

The two links are [hsw-cli](https://hsd-dev.org/api-docs/#getwalletinfo) and [hsw-rpc](https://hsd-dev.org/api-docs/#selectwallet) to show both ways to select wallet.

Issue #204, opened 2020-08-14
Add wallet rescan button to Settings


sdtsui:
Related to #203

Issue #202, opened 2020-08-14
Misclassified TX


chikeichan:
Some Reveals are getting classified as "Sent Funds" in Bob

PR #207, opened 2020-08-19
Rescan Button and various fixes

chikeichan:
- add a rescan button to Settings #204 
- fix parsing misclassified reveal tx #202 
- remove WalletSync modal and replace with Rescanning status in header. This allows users to still navigate in Bob when the wallet is rescanning 
Screen Shot 2020-08-19 at 3 33 28 PM
- added a loading transaction spinner to show that transactions are loading. This should be unnecessary when pagination is implemented since load time would be much shorter.
Screen Shot 2020-08-19 at 3 33 50 PM

### DIFF
--- a/app/mainWindow.js
+++ b/app/mainWindow.js
@@ -50,3 +50,7 @@ export default function showMainWindow() {
 export function getMainWindow() {
   return mainWindow;
 }
+
+export function dispatchToMainWindow(reduxAction) {
+  mainWindow.webContents.send('ipcToRedux', reduxAction);
+}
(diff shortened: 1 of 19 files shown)


app/mainWindow.js
```diff
@@ -50,3 +50,7 @@ export default function showMainWindow() {
 export function getMainWindow() {
   return mainWindow;
 }
+
+export function dispatchToMainWindow(reduxAction) {
```
chikeichan, line 54: want to highlight this new method to allow dispatching redux action from the main process in electron. The goal is to move wallet state management (such as backgroundMonitor.js) to the main process and optimize network activities between the wallet node and `hsd` node.

PR #200, opened 2020-08-13
Start with migrate change flag

pinheadmz:
See https://github.com/handshake-org/hsd/pull/415

and especially:

https://github.com/handshake-org/hsd/pull/415#issuecomment-670677057

and

https://github.com/handshake-org/hsd/pull/415/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1e

The next release of hsd will fix a bug that corrupted the hsd wallet if the user manually generated change addresses from the API (this will affect a grand total of about 3 people worldwide, probably). The bug is already fixed but repairing corrupt wallets is an issue closed by the PR above. We had trouble deciding how much to do "automatically", the process requires a backup, database migration and possible rescan. Since the repair process is opt-in its a bit tricky to do when hsd is embedded like in Bob. This PR is a demonstration on how it should be done. I changed the hsd dependency in this PR if developers want to test.

Once this and the hsd PR are merged, the UX will be one of the following:

- User starts up Bob after upgrade, hsd checks wallet for corruption (may take a minute or so)...
  -  No corruption found, life goes on. (the change in this PR can be eventually removed, it only needs to be called once)
  - Corruption found. Because the flag is set, the wallet regenerates the missing change addresses and then **initiates a rescan**

It's a bit tricky from the `hsdMain.js` file to get the UI-blocking "wallet rescan" modal from https://github.com/kyokan/bob-wallet/pull/117, so unfortunately at this point if the user is in the second group, they may not know their wallet is rescanning and the UI may show some unexpected data. Not sure how to address this besides adding notes in the Bob changelog and release notes? 

### How to test:

- fresh data dir (dev mode)
- change the `hsd` dependency in package.json to `"hsd": "https://github.com/handshake-org/hsd#v2.1.3",`
- generate a few change addrs with cli: `$ hsw-cli --id=allison --account=default change`
- let bob sync for a while, doesn't matter. then close
- change the `hsd` dependency in package.json to `"hsd": "https://github.com/handshake-org/hsd#change-migration",`
- restart bob. 

If you remove the patch I added to `hsdMain.js`, Bob will not load completely and an error will be thrown in the console window for hsd:

```
hsdMain.js:47 Error: Wallet is corrupted.
Back up wallet and then restart with
`hsd --wallet-migrate=0` or `hs-wallet --migrate=0`
(Full node required)
    at WalletDB.migrateChange (/Users/matthewzipkin/Desktop/work/bob-wallet/node_modules/hsd/lib/wallet/walletdb.js:262)
[...]

PR #232, opened 2020-10-14
upgrade hsd to rescan fix

chikeichan:
This PR builds on #231 and should be rebased against master after #231  is merged.

This upgrade `kyokan/hsd` to support:
- high lookahead value (from uint8 to uint32)
- fix bug in rescan where wallet with high lookahead value would break (or require multiple rescan attempt)

You can check out [this pull request to hsd](https://github.com/handshake-org/hsd/pull/519) for comments/review on the `hsd` part of the code.


chikeichan:
closing for now -- working on new version of `kyokan/hsd` that doesn't bump lookahead to uint32 -- will reopen after

chikeichan:
Ready to review! This version of `hsd` still include a fix for filter update during rescan, but no long change lookahead data type. Instead, it only hardcodes the value at `1000`.

app/ducks/walletActions.js
```diff
@@ -187,7 +187,13 @@ export const fetchTransactions = () => async (dispatch, getState) => {
     payload: true,
   });
 
-  const txs = await walletClient.getTransactionHistory();
+  let txs = [];
+
+  try {
+    txs = await walletClient.getTransactionHistory();
+  } catch (e) {
+    //
```
mslipper, line 195: Don't we want to do something with this error? Tx history will be empty if an error occurs here.

app/background/wallet/service.js
```diff
@@ -547,6 +550,7 @@ class WalletService {
       prefix: networkName === 'main'
         ? HSD_DATA_DIR
         : path.join(HSD_DATA_DIR, networkName),
+      migrate: 0,
```
mslipper, line 553: Won't this prevent any future wallet migrations from going through?

app/background/wallet/service.js
```diff
@@ -594,8 +596,15 @@ class WalletService {
...
     });
+  };
 
-    await this.node.wdb.addBlock(entry, txs);
+  onNewBlock = async (entry, txs) => {
+    await this._ensureClient();
+    await this.refreshNodeInfo();
+
+    if (entry && txs) {
+      await this.node.wdb.addBlock(entry, txs);
```
mslipper, line 606: Doesn't the wallet node do this automatically?

app/background/wallet/service.js
chikeichan: For more context here is the original [PR](https://github.com/kyokan/bob-wallet/pull/200) and intent before the `migrate` flag addition. It is for an existing migration in `hsd`.

app/background/wallet/service.js
chikeichan: Yep - i am stubbing the handler to call `refreshNodeInfo` on new block vs on an interval.

app/ducks/walletActions.js
chikeichan: i was trying to handle the error when wallet name is not set, which should return an empty array, but now I realize this is better handled in `wallet/service` instead. 

because there are many PRs stacked on this one now, I will fix this issue in my next commit.

chikeichan: e06c913

PR #233, opened 2020-10-15
Add multi wallet support

mslipper:


app/pages/Settings/index.js
```diff
@@ -207,10 +207,10 @@ export default class Settings extends Component {
           () => history.push('/settings/wallet/reveal-seed'),
         )}
         {this.renderSection(
-          'Remove wallet',
-          'This will remove all data from Bob. Proceed with caution.',
```
sdtsui, line 211: Not sure why but when I run this from `4fb4c1` - `Reset not remove`, I don't see the new copy.
Seems more likely it's local to me, but can you double check?  Is it possible that the Reset props are elsewhere in the codebase, and this new copy isn't actually being loaded?

Screen Shot 2020-10-16 at 12 17 07 AM

app/pages/Settings/index.js
```diff
@@ -314,15 +314,6 @@ export default class Settings extends Component {
         { this.renderContent() }
         <Switch>
           <Route path="/settings/wallet/account-index" component={AccountIndexModal} />
-          <Route
```
sdtsui, line 317: It looks like something changed about the way you're working with routes in settings - I'm getting a warning in console every time I execute that code path in Settings.  Details below:

slash-account

sdtsui (commented):
This might be breaking - I can no longer reveal seed phrase in this build.

Error in UI, and terminal output:
Screen Shot 2020-10-16 at 12 27 35 AM


Note: In order to run this branch, and aware of the fact that I did a full resync with Chikeichan's recent branch, I removed the directory at  `~/Library/Application Support/Electron/hsd_data/wallet` and to avoid incompatibility issues.  I also deleted `package-lock` and `node_modules` before `npm run dev` normally.

Let me know if I need to do more locally to test this properly.  Some of my comments are UI only so might be relevant either way.

sdtsui:
Q for @mslipper : is there no way to remove a wallet?  I checked `hsd-dev.org`, I guess the only way to do it is delete the wallet folder and re-add the relevant wallets manually?

pinheadmz:
@sdtsui wallet removal is [implemented in walletDB](https://github.com/handshake-org/hsd/blob/c85d9b4c743a9e1c9577d840e1bd20dee33473d3/lib/wallet/walletdb.js#L962-L986) but not exposed in the API because I think we're afraid users will destroy their money.

chikeichan:
Thanks @mslipper ! I have pulled and ran the branch with no issues. Seems like the two issues from @sdtsui are already resolved. My only comment is that i think we should default the account dropdown selector to `allison` if `allison` exists. This lowers confusion for existing users who is coming from an old version. wdyt?

@sdtsui we should include in our release note about the `allison` key for existing user upgrading.

@pinheadmz is the concern about the users not backing up their seed phrase + unrevealed bids? or do you mean that the wallet removal method could be unsafe to use (i.e. removing one wallet leading to loss of data form another)? Trying to see if this is something the warning modal will address, as we already have a "remove wallet" button with a warning modal. It would be nice if user can remove individual wallet.

chikeichan (approved):

chikeichan:
Here is an experimental [branch](https://github.com/kyokan/bob-wallet/compare/merge-multi-rescan?expand=1) with multi account support + rescan filter fix rebased on top for anyone interested in testing. I am currently running the branch to identify issues to work on prior to next release candidate.

PR #238, opened 2020-10-19
Hide unencrypted wallets

### LINKED ISSUE #236: wid `primary` should be password-protected
Related to multi-account #233 

The wid `primary` is automatically created on start up by `hsd` and is an unencrypted wallet, meaning that user can "login" with any password.

We should not allow users to use this wallet. I think a simple way to handle it would be:
- filter wallet ids by `encrypted === true` in `wallet/service.js`
- disallow `primary` from frontend
- default wid during onboarding to `main`

If we can remove wallet by id later, we can re-allow the `primary` wid by remove the existing `primary` before creating the `primary` wallet again

chikeichan:
partially fix #236 

This filters the accounts dropdown by the wallet's encryption status, hiding the `primary` wid effectively.

Also fixes a bug where new wallet doesn't show up on create/import.


mslipper:
Wouldn't we want the primary wallet to appear? If people use the CLI to manage the primary wallet it might be weird from a UX standpoint for it to be hidden in the UI.

chikeichan:
We do want the `primary` wallet to appear, but i think there are a few UX issues to work out first prior to adding it back. I added the details to #245 . I was thinking that hiding `primary` is okay because it's the same treatment as previous version of Bob, but i can see the argument that users would expect to see all wallets with a new **multi account** feature. What do you think?

PR #243, opened 2020-10-19
stub rescan handler + reset wallet state when switching wallet

### LINKED ISSUE #235: unclear rescan status when importing second wallet
In this [branch](https://github.com/kyokan/bob-wallet/compare/merge-multi-rescan?expand=1), header status stays at "Synchronized" or stucked at zero after a second wallet is imported, but the rescan is actually happening normally in background. 

TODO:
- [ ] it should show correct rescan status. We can stub the `wdb.on('block rescan')` handler in `wallet/service.js` to set wallet status to rescan whenever we receive a block from the **block rescan** event

chikeichan:
closes #235


app/background/wallet/service.js
```diff
@@ -592,11 +586,12 @@ class WalletService {
     });
 
     node.wdb.client.bind('block connect', this.onNewBlock);
+    node.wdb.client.unhook('block rescan');
```
pinheadmz, line 589: How does this affect the CLI command `rescan` from API / hs-client? It might be ok to replace that endpoint with this custom GUI action, but we should be sure that if a user "accidentally" initiates a rescan from outside Bob (directly to the wallet node HTTP API)  that it doesn't leave the wallet in some weird state.

Issue #188, opened 2020-07-31
URL Validation for CustomRPC Form


chikeichan:
- [x] CustomRPCModal should handle network type 
- [x] CustomRPCModal should validate that the endpoint has the right network type
- [x] CustomRPCModal should validate the url on frontend (parseable by the URL Class)
- [ ] Test Connection action should surface better error message for debugging (e.g. wrong network, api not found, etc)

-----------------
Fresh install, empty datadir, using `npm run dev`, loads OK.

 I love the new settings page layout, very clean and easy to navigate 👍 

But still having some issues.

Started fine, switched to regtest. Went to settings and clicked 'stop' -- bob logged me out and returned to "enter your seed phrase" screen again. I did so, and then I got stuck on "syncing wallet 0%" -- I didn't have a chance to enter the IP of my remote hsd node yet.

So quit and relaunched. This time I was able to go to settings and enter remote hsd info. I tried setting remote node url to a raspberry pi running hsd on my LAN by its IP address. I think there should be a timeout here because it just tried connecting forever and I must've entered somthing wrong. Couldn't cancel and try again.  I noticed this error in the log: ` Invalid URL: 10.0.1.5`

Quit and relaunch. This time stuck on "loading node...". Brick wall, every restart.

Cleaned out datadir and start over, try again. This time I set the remote node URL specifically to `http://10.0.1.5:12037` (because I also noticed in the log it was trying to hit port `80`, not the default rpc port, so without specifying port it didn't connect)

Now we're in business! I watched the remote hsd log, Bob seemed to rescanning OK - although I wasnt seeing the "syncing %" modal from https://github.com/kyokan/bob-wallet/pull/117  While rescanning, the Bob UI doesn't really work, it won't load auction details etc until the rescan job is done.

Before rescan was done though I tried something else, switching to another node. This time, local regtest. So set the URL to `http://127.0.0.1:14037` Bob didn't complain but it should have - Bob was configured for mainnet but that node is on regtest. I noticed the "network" setting is disabled when you configure a remote hsd -- I don't think it should be -- a mainnet wallet trying to rescan from a regtest node will be weirdo, and there should maybe be a check in place for that.

I'm happy to help with code review but the PR is really big (`+13,009 −160` !) It might be easier to get review if it were split up into a few smaller patches. Like
1- refactor settings page
2- use walletnode internally
3- option to switch to remote node
[...]

PR #254, opened 2020-10-28
allow custom rpc to switch network type + cache namehash to minimize request to node

chikeichan:
relates to #188


sdtsui:
proptypes fail when clicking import "import wallet" for the first time in dev

Screen Shot 2020-10-28 at 5 38 22 PM

sdtsui:
another one inbetween step 2 and 3 when creating pw
Screen Shot 2020-10-28 at 5 42 27 PM

sdtsui (commented):
@chikeichan  - QA complete.  Apart from the create issue discussed in person, and assorted proptypes 'onStart' (if you care), this is gtg to me.

@pinheadmz - if you have time, a once over would be helpful since you've been thinking about custom rpc stuff since the feature was conceived, `Roberto!` has it, etc. Thanks!

chikeichan:
@sdtsui both error message + the create new account bug is fixed 27de837

pinheadmz:
@sdtsui heres my hack to read port assignments from ENV: https://github.com/pinheadmz/bob-wallet/commit/6f730d60b8c2c3911c995c012d24467a863bcdc7

I'm using the same params as hsd itself, could be better. What would be really great is if we could access a config file as well. It's a little tricky because hsd doesn't run in the shell, but maybe Bob could require [bcfg](https://github.com/bcoin-org/bcfg) and then pass the parsed options to hsd internally.

PR #266, opened 2020-11-03
Ledger support

mslipper:
Screen Shot 2020-11-03 at 1 21 31 AM


app/pages/AcountLogin/index.js
```diff
@@ -64,18 +67,20 @@ export default class AccountLogin extends Component {
...
-              })}
-              type="password"
-              placeholder="Your password"
-              onChange={e =>
-                this.setState({passphrase: e.target.value, showError: false})
-              }
-              value={passphrase}
-              autoFocus
-            />
+            {currWallet !== 'Ledger' && (
```
pinheadmz, line 70: hm, interesting - so a Ledger wallet can't be locked for privacy? I wonder if we can add a passphrase to the unused `primary` wallet in this case just to have something? Actually, I believe a watch-only wallet can still be locked / unlocked. Even watch only wallets have a master private key that can be encrypted. The key isn't used for addresses but it is used for generating auth tokens.

chikeichan:
replace by #344

Issue #217, opened 2020-09-07
Wallet displays incorrect output - does not match with addresses controlled


sdtsui:
![Screen Shot 2020-09-07 at 3 24 19 PM](https://user-images.githubusercontent.com/8230144/92418852-ace19800-f11e-11ea-94aa-301905062683.png)
https://hnscan.com/tx/566db20588cef3420ac218334d871a6feb5cdbd3dfb3708428bbea6ee4449349
![Screen Shot 2020-09-07 at 3 24 27 PM](https://user-images.githubusercontent.com/8230144/92418837-920f2380-f11e-11ea-9d23-a87d859180f1.png)

I believe this is a known issue. If anyone (@pinheadmz maybe?) knows where this is being addressed in HSD, please let me know or mention it in a comment.  The displayed output should be shown in tx history as `Airdrop Claim` not `Mining Reward`.

OS: macOS 10.14.6
Bob Version: 0.5.0 (Stable)

pinheadmz:
The per-transaction value and label comes from [this function](https://github.com/kyokan/bob-wallet/blob/0c89d808188003ce4438ad7317c8d7809d458231/app/ducks/walletActions.js#L268-L359) in Bob, not hsd. Airdrop claims are actually coinbase transaction outputs. Bob could have extra logic to attempt to differentiate the two (it would require examining the corresponding input for the output in question, and decoding the blob). I have a crude method for differentiating between airdrop, faucet, claim, and block subsidy payouts:  https://gist.github.com/pinheadmz/a2cd8d6d32ae6c5fab75eb41497f070a#file-hsd-cov-stats-js-L32-L41

pinheadmz:
Just did a quick test on latest master and I think something got buggy here. I opened bob in regtest and sent it 100 HNS from a second regtest node that had mined block. The total balance was correct in Bob (100 HNS) but the portfolio screen said the incoming transaction was for `+2000.000000`. SO that's weird, that is probably the value of the *input* of the tx it received (from the miner)...

pinheadmz:
I would double check the logic here: https://github.com/kyokan/bob-wallet/pull/207/commits/7bbe70cc00f2e6c22b0ce93d58550fad547455fe

rithvikvibhu:
Fixed in #268.

PR #268, opened 2020-11-06
ui: fix transaction output for coinbase tx

chikeichan:
related to #217 

this fixes the issues where transaction is not showing the correct _value_ -- it does not fix the _categorization_ issue (i.e. **Airdrop** being displayed as **Mining Reward**)


app/ducks/walletActions.js
```diff
@@ -332,7 +332,9 @@ async function parseInputsOutputs(net, tx) {
 
     // Track normal receive amounts for later
     if (covenant.action === 'NONE') {
-      totalValue += output.value;
+      if (await walletClient.hasAddress(output.address)) {
```
pinheadmz, line 335: Isn't this the same as just checking if `output.path` is non-null?

chikeichan: didn't know there is a `output.path` variable 😄  

fixed now! I also fix `wallet/service` to inflate transactions before returning.

sdtsui (approved):
Screen Shot 2020-11-09 at 12 19 57 AM

Getting this error when switching to mainnet from simnet, but I don't think it relates. Otherwise looks fine.  I suppose the plan is to do categorization another time?  I think folks will be happy with any improvement here.

chikeichan:
I will checkout the error but pretty sure it's unrelated. And yep - i will fix categorization in another PR

Issue #252, opened 2020-10-26
Required password for confirming bid


Falci:
Go to the bid page, place a bid. 
Bob will ask to mark the check box. After confirming, Bob will ask the password.

![image](https://user-images.githubusercontent.com/1330573/97148453-77931700-176b-11eb-977a-7c013de36b35.png)

I'm assuming it refers to the passphrase used to open Bob.
Bob will accept and proceed with whatever password you put there

## Expected Behavior
Reject wrong passwords

## Context (Environment)
Bob 0.5.0

pinheadmz:
Bob needs your passphrase any time it signs a transaction. When you put in the passphrase, the wallet is decrypted in memory (the master private key is anyway) for 60 seconds. Is it possible that in your case you had recently put in the passphrase less than 60 seconds ago? Perhaps you logged in to the wallet then placed the bid within one minute? If so, the wallet was probably already decrypted and so the TX could be signed -- effectively the password dialog was redundant.

We ask the user for the passphrase every time a TX is signed just in case its been longer than 60 seconds since you logged in:
https://github.com/kyokan/bob-wallet/pull/128

Falci:
>  Is it possible that in your case you had recently put in the passphrase less than 60 seconds ago?

Pretty sure this is the case.

> effectively the password dialog was redundant.

It may not be a hsd requirement, but I would keep the validation and, well, validate it. Just to be sure a malicious user didn't perform an action if I forget it open.

pinheadmz:
Bob doesn't really validate the password, hsd tries to decrypt the wallet with it. If the wallet is already decrypted the subsequent unlock call just extends the amount of time the private key is held in memory:

https://github.com/handshake-org/hsd/blob/c85d9b4c743a9e1c9577d840e1bd20dee33473d3/lib/wallet/masterkey.js#L156-L163



The best approach would be to go around every place the private key is used in bob and lock the wallet as soon as signing is done - so we can always assume the application is safe from malicious users ;-)

PR #272, opened 2020-11-21
wallet: fix missing return value when sending tx

chikeichan:


app/background/wallet/service.js
```diff
@@ -412,14 +412,15 @@ class WalletService {
   send = (to, amount, fee) => this._ledgerProxy(
     () => this._executeRPC('createsendtoaddress', [to, Number(amount), '', '', false, 'default']),
     async () => {
-      await this.client.send(this.name, {
+      const res = await this.client.send(this.name, {
```
chikeichan, line 415: This response was removed in #252 .

To prevent future regression, let's start to add unit tests around critical code path #273

PR #274, opened 2020-11-22
feature: add deeplink capability to bob

chikeichan:
This PR allow others to deeplink to bob using `bob://` protocol name. 

The two deeplink URLs that are implemented now are: 

- `bob://openname?name=test2` -- open auction page
- `bob://openmanager?name=test2` -- open domain manager page
- `bob://updaterecord?name=test2&txt=hello` -- 1-click update record

**Expected Deeplink Behavior**
- when opening a deeplink on a *locked* wallet, bob will allow user to choose a wallet, login, then redirect to the deeplink
- when opening a deeplink on an *unlocked* wallet, bob will redirect to the deeplink

Also as part of the PR, Domain Manager view is updated to disable all inputs/action when user is not the owner.

deeplink

deeplink2


pinheadmz:
Interesting. As long as you're playing around with this, maybe Bob could support [BIP21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)-style URIs for payments as well. That will also be more generic and can be standardized across the ecosystem with our own ["HIP"](https://github.com/handshake-org/HIPs)

example:

```
handshake://hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x?amount=1.123456
```

chikeichan:
> Interesting. As long as you're playing around with this, maybe Bob could support [BIP21](https://github.com/bitcoin/bips/blob/master/bip-0021.mediawiki)-style URIs for payments as well. That will also be more generic and can be standardized across the ecosystem with our own ["HIP"](https://github.com/handshake-org/HIPs)
> 
> example:
> 
> ```
> handshake://hs1qde7jaw6qgzzfu83upn3twvsyhh0zrshg76qe0x?amount=1.123456
> ```

I did not know that there is a BIP for URI schemas -- thanks!!

For this particular use case I am just looking to unblock a way to deeplink Bob to the Domain Info and DNS Record screen. Since the use case is fairly limited right now, i will keep this Bob-specific for now, and will keep an eye out for timing to submit/integrate to Handshake's own version of BIP21.

app/index.js
```diff
@@ -19,6 +23,69 @@ ipcRenderer.on('ipcToRedux', (_, message) => {
...
+        store.dispatch(setDeeplinkParams({ txt }));
+      }
+
+      if (isLocked) {
+        store.dispatch(setDeeplink(message));
+      } else {
+        history.push(`/domain_manager/${name}`);
+      }
+      return;
+    case "//openauction":
```
sdtsui, line 56: Would it make more sense to have `openauction` be `openname`, to match more closely with the ecosystem's block explorers?   i.e. The on-chain auction/ownership page is the high level view of the name, so you're opening 
 (information for) the name.

Same line of thinking: `opendomain` could be `openmanager` or `namemanager` as name itself would be implied by a param.  I'm not sure if you'd prefer verb-prefixes, so maybe `openmanager` is best.

app/main.js
```diff
@@ -24,6 +24,14 @@ if (
   require('electron-debug')();
 }
 
+app.setAsDefaultProtocolClient('bob');
+
+// Protocol handler for osx
```
sdtsui, line 29: Is this osx only for now?  Ping me on Signal and I can test Windows whenever it's ready.

sdtsui (commented):
Main flows QA'd with no issues.  Left two questions @chikeichan .

app/index.js
chikeichan: updated

app/main.js
chikeichan: just added support for window as well

Issue #201, opened 2020-08-14
Dial to select block explorer in Settings


sdtsui:
@chikeichan

sdtsui:
Users want to be able to use block explorers other than HNScan.  Note that the URL formats may be different for txs.
I'm thinking we start with Shakescan so there's at least a format for others to contribute.
https://shakescan.com/

pinheadmz:
duplicate? https://github.com/kyokan/bob-wallet/issues/152

Issue #152, opened 2020-06-28
[Feature] HNScan alternative


Falci:
I had a look on the HNScan project and decided to run it on my node.
I'd like to configure it on Bob. When I click in a transaction, it should open in my custom HNScan.

sdtsui:
Thanks for the suggestion @Falci .  Clarifying: you would _not_ like to be redirected to the website at hnscan.com, but your own locally hosted version?  Is this accurate?

If so, can you tell me a little more about why?  Just curious if it's for security features, if you're building something special, or anything else. :)
(Feel free to DM on Telegram if you prefer!)

Falci:
> Clarifying: you would not like to be redirected to the website at hnscan.com, but your own locally hosted version? Is this accurate?
Yes, exactly.

Not a big deal, but running hnscan in my node is faster than the hnscan.com.

pinheadmz:
This would be a cool setting. I've seen implementations in other Bitcoin applications, and basically there is just a text input field in settings where you can insert any URL and a `%s` placeholder for the txid. So for example, these would all be valid settings:

`https://hnscan.com/tx/%s`

`https://shakescan.com/transaction/%s`

`http://127.0.0.1:8080/tx/%s`

brandondees:
Yeah I like this idea as well. hnscan.com is down / dead sometimes, and I sometimes prefer shakescan.com, or might want to run my own customized thing.

Tangentially related, it'd be extra cool to be able to register bob as a url handler for `hns://` or something so that other tools can link _into_ bob views as well. This would open the door to a lot of easy interop with other HNS tools (I'm specifically imagining exporting a namebase watchlist into bob with a single click, or navigating from any block explorer to the Bob bidding view for a name, or pre-populating HNS transfer details specified by the requester, etc.)

PR #282, opened 2020-12-16
Use hnsnetwork.com

Falci:
Replace the hnscan.com by hnsnetwork.com

Reason: hnsnetwork.com seems more stable.

Right now, hnscan.com is stuck with 1 day block delay


pinheadmz:
see #152 #201 I agree hnsnetwork is better but the best possible fix is expose it as a setting, then users can switch whenever one explorer goes down very easily:

https://github.com/kyokan/bob-wallet/issues/152#issuecomment-662117108

Falci:
Consider it a temporary fix while #152/#201 is not ready.

chikeichan:
duplicate of #285

PR #285, opened 2021-01-14
Explorer selection

### LINKED ISSUE #152: [Feature] HNScan alternative

### LINKED ISSUE #201: Dial to select block explorer in Settings

rithvikvibhu:
Closes #201, Closes #152.

This PR adds a dropdown selection for selecting the explorer used to open transactions.
All known explorers are listed, defaults to HNScan.

image
image

## Future work
- Opening names on explorers
- Custom explorers with URL templates


rithvikvibhu:
Currently, changing the explorer works and the transactions open as expected. But this change doesn't persist when Bob closes and open again. I've looked at how custom RPC is stored and retrieved, but not sure where the code for storing explorer selection should go.

Maybe in `app/background/node/service.js`? This is not strictly a NodeService kind of field though. Or do you think a new file for general settings would be better?

chikeichan:
> Currently, changing the explorer works and the transactions open as expected. But this change doesn't persist when Bob closes and open again. I've looked at how custom RPC is stored and retrieved, but not sure where the code for storing explorer selection should go.
> 
> Maybe in `app/background/node/service.js`? This is not strictly a NodeService kind of field though. Or do you think a new file for general settings would be better?

Thanks for the PR @rithvikvibhu !

I think writing those in `app/background/setting/service` would make sense -- there are likely other things we'd like to persist like you suggest.

rithvikvibhu:
>I think writing those in app/background/setting/service would make sense

Done, and it seems to work. Let me know if anything needs to change.

rithvikvibhu:
Also, how do you feel about this link for the name too?
image

pinheadmz:
My 2 cents: hnscan is not as reliable as it once was and is not under as regular maintenance as hnsnetwork.com

rithvikvibhu:
Hm. Wanna change the default to another one (if so, which)?

pinheadmz:
> Hm. Wanna change the default to another one (if so, which)?

http://hnsnetwork.com/ is my suggestion but as long as there is a way to choose I'm not picky

chikeichan:
reviewed + qa all block explorers. Also pushed a commit to update default to HNSNetwork.

LGTM!

chikeichan (approved):

PR #286, opened 2021-01-14
auction: close bidding 1 block early

### LINKED ISSUE #278: Off-by-one problem between bid end and reveal start
Bob allows bidding even on the block right after the bidding period ends. It fails with an error for hsd.

Example: `fairies/` (pic below for ref)
- At current height 43522: Allowed to bid, works
- At current height 43523: Not allowed to bid, but still shows bid button (but errors out as expected)
- At current height 43524: Not allowed to bid, doesn't allow as expected

@pinheadmz explains it well:
>What happens is, if the NEXT block is the first REVEAL block, that means no more bids are allowed. But according to hsd the name is still in a BIDDING state until that block is mined. What bob needs to do is check if bids/reveals are still allowed in the NEXT block. And in fact the hsd wallet does this.

## Expected Behavior
The bid button should disappear 1 block early.

## Screencap
image

rithvikvibhu:
This fixes #278.

The changes made checks if `blocksUntilReveal` is exactly 1 and overrides `isBidding()` and `isReveal()`, effectively closing the bidding (and hiding the Bid button) early.
The Bidding Ends time is also reduced by 1 block.

This PR also fixes:
- Label on the right (when bidding) from **Reveal Ends** to **Bidding Ends**
- **Your Bids** to **Bids** as all bids are shown

## Example
`y/` and `z/` are opened 1 block apart at heights 109 and 110.

At height 119:
```
y/:
bidPeriodStart: 115
bidPeriodEnd: 120
blocksUntilReveal: 1

z/:
bidPeriodStart: 116
bidPeriodEnd: 121
blocksUntilReveal: 2
```
image
image

### DIFF
--- a/app/components/AuctionGraph/index.js
+++ b/app/components/AuctionGraph/index.js
@@ -142,7 +142,7 @@ export default class AuctionGraph extends Component {
           <div className="auction-graph__column__headline">Reveal Period</div>
           <ProgressBar percentage={revealProgress}/>
           <div className="auction-graph__column__text__row">
-            { this.maybeRenderDateBlock(() => isBidding(domain), stats.bidPeriodEnd, stats.hoursUntilReveal) || this.maybeRenderDateBlock(() => isReveal(domain), stats.revealPeriodStart, stats.hoursUntilClose) || this.renderPlaceholder(REVEALING < currentStep)}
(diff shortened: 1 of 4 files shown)


chikeichan:
Thanks for all the PRs @rithvikvibhu ! This one LGTM as well

chikeichan (approved):

PR #288, opened 2021-01-15
my-domain: only show possible actions

### LINKED ISSUE #147: Confusion around "register"
Some errors tell users they must register their name, but Bob doesn't literally have any Register buttons -- maybe "manage my domain" should say "Register my domain" if it hasn't been registered yet?

### LINKED ISSUE #185: Renewal button/link is active when it's not permitted
## Expected Behavior

When I win an auction, bob should not show actions like renew when I cannot perform them, or show them as disabled actions.

## Current Behavior

The UI shows a 'Send Renewal' link on the domain manager page for names I haven't redeemed yet. If I click it, I get an error "name must be registered". 
The UI also shows a big blue "Renew my Domain" button on the regular domain view for unredeemed names, but if I click that, it says "Must wait to renew" which just leaves me with questions about how long/why/etc.

## Steps to Reproduce

1. Win an auction
1. Wait for reveal to end
1. Look up the name in the domain manager view or regular domain view
1. Try to use the visible 'renew' action
1. Be disappointed and confused

## Screencap

image
image

rithvikvibhu:
Closes #185, Closes #147.

This PR:
- Shows **Send Register** instead of **Send Renewal** in My Domain when the name is not registered
- Does not allow transfer when the name is not registered
- Change label **Your Bids** to **Bids** in My Domain as all bids are shown
- Show **Register my domain** instead of **Renew my domain** on Auctions page when the name is not registered

image
image

### DIFF
--- a/app/pages/Auction/BidActionPanel/Owned.js
+++ b/app/pages/Auction/BidActionPanel/Owned.js
@@ -19,6 +19,7 @@ class Owned extends Component {
     chain: PropTypes.object.isRequired,
     name: PropTypes.string.isRequired,
     sendRenewal: PropTypes.func.isRequired,
+    sendRegister: PropTypes.func.isRequired,
     showSuccess: PropTypes.func.isRequired,
     showError: PropTypes.func.isRequired,
     history: PropTypes.shape({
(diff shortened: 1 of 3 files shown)


rithvikvibhu:
Also including #147 because it's similar.

When a name is won (and the name isn't registered), the Auction page now shows a Register button instead of Renew.
After registering, it switches to Renew.

Kept the manage button in place because it lets you go the Domain Manager page to update records (alternative to Register button).

image
image

chikeichan (approved):

PR #289, opened 2021-01-15
my-domain: show next tree update under Records

### LINKED ISSUE #187: [Feature] Track next tree update from manage domain view
I registered some won domains in bob today, and when adding dns records, I noticed I don't have an easy way to guess when the records will be "live" on the chain, which I understand will occur at the next tree update interval. It'd be really handy for UX if the manage page would show what block is the next tree update and estimated time till then.

rithvikvibhu:
Closes #187.

This PR adds the next tree update block and time estimate above the Records table.

image

### DIFF
--- a/app/components/Records/index.js
+++ b/app/components/Records/index.js
@@ -39,18 +39,6 @@ class Records extends Component {
     return !deepEqual(this.props, nextProps) || !deepEqual(this.state, nextState);
   }
 
-  static renderHeaders() {
-    return (
-      <HeaderRow>
-        <HeaderItem>
-          <div>Type</div>
-        </HeaderItem>
-        <HeaderItem>Value</HeaderItem>
-        <HeaderItem />
-      </HeaderRow>
-    );
-  }
-
   constructor(props) {
     super(props);
     this.state = {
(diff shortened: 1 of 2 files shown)


rithvikvibhu:
Do you think this should be with context of the name's last update? Like "The records you just updated will be committed in X blocks" or "The changes you've made are live now".
Not sure how to get the last UPDATE tx of a name, so now it's always shown like in the pic.

chikeichan:
> Do you think this should be with context of the name's last update? Like "The records you just updated will be committed in X blocks" or "The changes you've made are live now".
> Not sure how to get the last UPDATE tx of a name, so now it's always shown like in the pic.

Making this in context would be nice - tracking on #293

For finding last UPDATE tx of any name, here is what I would do. Tagging @pinheadmz in case he knows of a better method.
- use [getnameinfo](https://hsd-dev.org/api-docs/#getnameinfo) to get `owner.hash` and `owner.index`
- use [gettxout](https://hsd-dev.org/api-docs/#gettxout) to get `address.string`, which is the owner's address. Documentation doesn't show that parameter `string` but it is there -- if this changes we can use the [Address](https://github.com/handshake-org/hsd/blob/master/lib/primitives/address.js) class in hsd to turn `hash` to `string`
- use [Get tx by address](https://hsd-dev.org/api-docs/#get-tx-by-address) to get all txs from this address. It should only contains TXs related to this name. If you are not testing this API on Bob, you need a node with `--index-address` option set to true.
- Iterate over TXs to find latest UPDATE of a particular name

chikeichan (approved):

pinheadmz:
You shouldn't need address indexer or tx indexer for this, since `owner` is always an unspent output, and therefore always available in the utxo set of any full node:

call `getnameinfo`  and get `owner` -- which is the UTXO that currently owns the name. In most cases this will be a REGISTER or UPDATE output, unless the most recent action is TRANSFER or RENEW.

Look up the utxo with https://hsd-dev.org/api-docs/#get-coin-by-outpoint height will be right there ;-)

chikeichan:
rebased and moved update text to table header

Screen Shot 2021-01-23 at 9 54 35 PM

Issue #144, opened 2020-06-14
[Bug] - Bid Dates always a single value


sdtsui:
I've checked 3 auctions that are still in reveal: `inan`, `ihn`, and `0i`.

Dates always look like this: `01/19/70`
![Screen Shot 2020-06-13 at 7 17 20 PM](https://user-images.githubusercontent.com/8230144/84583170-925f8300-adaa-11ea-9bc6-0fdfb80290ec.png)

panpansh:
same and more : 
![image](https://user-images.githubusercontent.com/790586/94344315-b3c54180-001e-11eb-8c2a-d69cec5ba843.png)

rithvikvibhu:
The main issue seems to have been fixed a while ago. And the second issue fixed in 290

PR #290, opened 2021-01-15
Auctions UI fixes

### LINKED ISSUE #93: [Copy Change] "Mask" should say "Bid + Mask"?
Screen Shot 2020-02-17 at 6 19 10 PM

I believe the last column on the right should say "Bid + Mask" not just Mask - I've placed a few bids now and just from looking at the transactions, it looks like either this column should say "Bid + Mask", or we should load subtract the bid amount before displaying.

WDYT, @mslipper @chikeichan ?

Please check on your end with bids you've placed: if this turns out to be a one-time bug, I'll try a few  more cases to repro.

### LINKED ISSUE #119: Bids list on auction screen should overflow:scroll
Eeek.
Screen Shot 2020-04-07 at 11 40 26 AM

### LINKED ISSUE #160: [Feature] See auction after domain registered
For some reason, Bob doesn't show the auction information (bids & value) after the domain is registered (not sure why/when).

It would be interesting if that information could still be accessed, forever.

image

rithvikvibhu:
(sorry, a bit bigger than I expected. Kept making small changes and ended up this way.)

### Changes in this PR
- Scrollbar appears only when bid list is longer than max-height (closes #119)
- Updated text (closes #93): Mask -> Lockup; Disguise -> Blind
- Show bids and records when appropriate (closes #160) - table for when exactly below
- Fixed Reveal Ends time (use `hoursUntilClose` instead of `stats.revealPeriodEnd` before, because this is block number, not time) - Doesn't show 7000+ days as reveal end anymore (https://github.com/kyokan/bob-wallet/issues/144#issuecomment-699511175)
- Update auction page (progress bars, etc.) as and when current block increases
- Fixed red warning when reveal period has just started - see screenshots 1 and 2
- Fixed Highest Lockup amount (replaced `domain.info.highest` with `Math.max(bids)`) + used displayBalance() for formatting - see screenshots 1 and 2
- Only attempt to show records after domain info is available (Loading...) - otherwise shows error in console

**Screenshot 1:** Old
image

**Screenshot 2:** New (at reveal period start block)
image

**Screenshot 3:** New (during reveal period)
image

#### Collapsibles shown on auction page at different states:
- New name without auction - Bids, Vickrey process
- During auction (opening, bidding, revealing): Bids, Vickrey process
- Closed state after auction (has domain info): Bids, Records

### DIFF
--- a/app/components/AuctionGraph/index.js
+++ b/app/components/AuctionGraph/index.js
@@ -41,6 +41,12 @@ export default class AuctionGraph extends Component {
     await this.getStatusInfo(this.props.domain);
   }
 
+  async componentDidUpdate(prevProps, prevState) {
+    if (this.props.chain.height !== prevProps.chain.height) {
+      await this.getStatusInfo(this.props.domain);
+    }
+  }
+
   getStatusInfo(domain) {
     const stats = domain.info && domain.info.stats || {};
 
--- a/app/components/Collapsible/collapsible.scss
+++ b/app/components/Collapsible/collapsible.scss
@@ -36,7 +36,8 @@
     opacity: 1;
     // padding: 1rem;
     position: relative;
-    max-height: 50rem;
+    max-height: 35rem;
+    overflow-y: auto;
     height: auto;
     transition: 0.5s ease-out;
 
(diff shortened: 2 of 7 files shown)


rithvikvibhu:
Updated description with all changes, tried to keep each point to a commit.

For #93, I went with the changes suggested in the issue, even if there's no recent comments, because it makes sense imo. If it (or anything else) needs to change, lemme know.

PR #294, opened 2021-01-24
hsd: bump hsd to 0.2.3

chikeichan:
This PR updates bob to [this branch](https://github.com/kyokan/hsd/tree/kma2.3) of `kyokan/hsd`, which is rebased update to [this commit](https://github.com/handshake-org/hsd/commit/1a086e44b5b60e1ebb01955b0e1b2a0ebd00b42d) in `hsd`

### DIFF
--- a/package.json
+++ b/package.json
@@ -141,7 +141,7 @@
     "electron-updater": "4.1.2",
     "history": "4.7.2",
     "hs-client": "https://github.com/handshake-org/hs-client.git",
-    "hsd": "https://github.com/kyokan/hsd/tarball/1e1292d",
+    "hsd": "https://github.com/kyokan/hsd/tarball/82dd02b",
     "isomorphic-fetch": "^2.2.1",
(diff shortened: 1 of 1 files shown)


pinheadmz:
ok awesome, lets wait until https://github.com/handshake-org/hsd/pull/538 is merged into hsd as well so the official version number gets included in the package.json and therefore the peer metadata.

Also had another thought about this. What do you think about changing the `pkg` name in Bob:

https://github.com/kyokan/hsd/blob/kma2.3/lib/pkg.js#L17

to `bob-hsd` or something?

Then in `getpeerinfo` it will be revealed:
```
...
 "subver": "/bob-hsd:0.0.0/",
...
```

I know this is a privacy leak but I think it will be helpful in network analysis if anything goes wrong at the p2p or consensus layer. Bob is kinda an alternative implementation of hsd, it might be good to know that.

chikeichan:
closing for #301

PR #296, opened 2021-01-24
Fix tx value for atomic name swap finalize

### LINKED ISSUE #292: Atomic swap feature not showing received funds as a transaction
I just did an atomic swap as a seller and although I did receive the HNS (my wallet balance updated) I do not see it as a separate received transaction in the wallet. I only see the "Finalized" transaction. I checked with another seller that used the feature and the same happened to him.
Capture

rithvikvibhu:
Fixes #292.

The value for atomic name swaps show up as 0. This PR shows the +ve/-ve amount for seller/buyer.

**Note:** I could only test with 2 name swaps, probably needs more testing. Also not sure if this is the best way to get the values, it's the shortest I could think of.

For the example below, 2 names transferred, `c/` for **50 HNS**, `d/` **without payment**.

Sender:
image

Receiver:
image

### DIFF
--- a/app/ducks/walletActions.js
+++ b/app/ducks/walletActions.js
@@ -383,12 +383,29 @@ async function parseInputsOutputs(net, tx) {
       covValue += output.value;
     }
 
+    if (covenant.action == 'FINALIZE') {
+      const isSender = output.path == null;
+
+      // Start with no payment for normal Finalizes
+      covValue = 0;
+
+      // containsOtherOutputs: Look for outputs that are not change addresses to self
+      if (isSender) {
+        // If yes, sum all self outputs (payment made by receiver)
+        const containsOtherOutputs = tx.outputs.findIndex(x => x.covenant.action == 'NONE' && x.path === null) !== -1;
+        if (containsOtherOutputs) covValue = tx.outputs.reduce((sum, op) => op.path ? (sum+op.value) : sum, 0);
+      } else {
+        // If yes, sum all outputs to others' addresses (payment made to sender)
+        const containsOtherOutputs = tx.outputs.findIndex(x => x.covenant.action == 'NONE' && x.path !== null) !== -1;
+        if (containsOtherOutputs) covValue = -tx.outputs.reduce((sum, op) => !op.path ? (sum+op.value) : sum, 0);
+      }      
+    }
+
     // Renewals and Updates have a value, but it doesn't
     // affect the spendable balance of the wallet.
     if (covenant.action === 'RENEW' ||
       covenant.action === 'UPDATE' ||
-      covenant.action === 'TRANSFER' ||
-      covenant.action === 'FINALIZE') {
+      covenant.action === 'TRANSFER') {
       covValue = 0;
     }
 
--- a/app/components/Transactions/Transaction/index.js
+++ b/app/components/Transactions/Transaction/index.js
@@ -59,18 +59,20 @@ class Transaction extends Component {
           || tx.type === COINBASE
           || tx.type === REVEAL
           || tx.type === REDEEM
-          || tx.type === REGISTER)
+          || tx.type === REGISTER
+          || (tx.type === FINALIZE && tx.value > 0))
         && !tx.pending,
       'transaction__number--neutral':
         (tx.type === UPDATE
           || tx.type === RENEW
           || tx.type === OPEN
           || tx.type === TRANSFER
-          || tx.type === FINALIZE)
+          || (tx.type === FINALIZE && tx.value === 0))
         && !tx.pending,
       'transaction__number--negative':
         (tx.type === SEND
-          || tx.type === BID)
+          || tx.type === BID
+          || (tx.type === FINALIZE && tx.value < 0))
         && !tx.pending,
     });
 
@@ -129,6 +131,8 @@ class Transaction extends Component {
       content = this.formatDomain(tx.meta.domain);
     } else if (tx.type === 'FINALIZE') {
       description = 'Finalized Domain';
+      if (tx.value > 0) description = 'Received Payment for Domain';
+      if (tx.value < 0) description = 'Finalized With Payment';
       content = this.formatDomain(tx.meta.domain);
     } else {
       description = 'Unknown Transaction';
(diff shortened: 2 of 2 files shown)


app/ducks/walletActions.js
```diff
@@ -368,12 +368,31 @@ async function parseInputsOutputs(net, tx) {
       covValue += output.value;
     }
 
+    if (covenant.action == 'FINALIZE') {
+
+      // isSender if there exists an input with value 0 and own address
+      const isSender = tx.inputs.findIndex(x => x.value === 0 && x.path !== null) !== -1;
```
pinheadmz, line 374: Why `value === 0` here? TRANSFER and FINALIZE outputs may have a value (the second highest bid, the amount of the "burned" coins by auction winner). So if I am the original name owner, the finalize will have one input from my wallet address (the TRANSFER) which will have a `value >= 0` depending on the auction. I think you may have to look for a TRANSFER covenant in the input and check if its a wallet address (`path !== null`)

pinheadmz:
I think it'd be awesome if you could actually expand this list to include something like `Received Payment for Finalize` and `Finalized With Payment` 

https://github.com/kyokan/bob-wallet/blob/abdcdc88d5e27c83521a2ab8d950269e7c24355c/app/components/Transactions/Transaction/index.js#L131-L133

app/ducks/walletActions.js
```diff
@@ -368,12 +368,31 @@ async function parseInputsOutputs(net, tx) {
...
+      }
+    }
+
     // Renewals and Updates have a value, but it doesn't
     // affect the spendable balance of the wallet.
     if (covenant.action === 'RENEW' ||
       covenant.action === 'UPDATE' ||
-      covenant.action === 'TRANSFER' ||
-      covenant.action === 'FINALIZE') {
+      covenant.action === 'TRANSFER') {
```
pinheadmz, line 395: I could be wrong but I think if the wallet is receiving payment for a FINALIZE, then the payment value might just be equal to `totalValue` calculated above. I'm not sure if that'll work for the buyer though (who is sending money) but might save you some of the logic in the blob L371-390

app/ducks/walletActions.js
rithvikvibhu: Oh I see, hadn't thought of that. The inputs in the tx object only has address and path, and no covenants so did this to identify the input with the name.
To test this, I transferred a name that was finalized with payment so it's like `TRANSFER` -> `FINALIZE(pay)` and now `TRANSFER` -> `FINALIZE(pay)` again.

**2nd transfer** (finalized with 30 HNS)
Sender side:
```
inputs: Array(2)
0: {value: 0, address: "rs1q084zn6eawcvrhkt8v2agrywwwfc5u58flx9q3j", path: {…}}
1: {value: 0, address: "rs1qhhu0qhfwx8kl9qvprqn274w6v58msp490dy8ug", path: null}

outputs: Array(3)
0: {value: 0, address: "rs1q5uy4kg8ts6z5qzlmaazqywr4hrdv3tkz725v86", covenant: {…}, path: null}
1: {value: 19981740, address: "rs1qemjhwa032430m6e4wss0ycvrehytfk0j4jqkl6", covenant: {…}, path: null}
2: {value: 30000000, address: "rs1q6yrp0tvksy8espkjwcfwxc845sk0wdtyuh6ceu", covenant: {…}, path: {…}}
```

Receiver side:
```
inputs: Array(2)
0: {value: 0, address: "rs1q084zn6eawcvrhkt8v2agrywwwfc5u58flx9q3j", path: null}
1: {value: 49988200, address: "rs1qhhu0qhfwx8kl9qvprqn274w6v58msp490dy8ug", path: {…}}

outputs: Array(3)
0: {value: 0, address: "rs1q5uy4kg8ts6z5qzlmaazqywr4hrdv3tkz725v86", covenant: {…}, path: {…}}
1: {value: 19981740, address: "rs1qemjhwa032430m6e4wss0ycvrehytfk0j4jqkl6", covenant: {…}, path: {…}}
2: {value: 30000000, address: "rs1q6yrp0tvksy8espkjwcfwxc845sk0wdtyuh6ceu", covenant: {…}, path: null}
```

The name input value is always 0, and the values are calculated and shown as +/- 30 HNS properly.
I guess this works because there's no value for `TRANSFER`s, and `FINALIZE` can only happen after a `TRANSFER`.

app/ducks/walletActions.js
rithvikvibhu: Yup, this works for the sender and I initially tried this (but outside the for loop because totalValue won't be complete inside the loop).

```
if (covAction) {
    return {
      ...covData,
      fee: tx.fee,
      value: covenant.action === 'FINALIZE' ? totalValue : covValue, // <- instead of just covValue
    };
  }
```

Like you said, this only works for the seller, not buyer. There's also a problem of identifying if the value is +ve or -ve.

app/ducks/walletActions.js
pinheadmz: Are you testing the swaps with a name you "won" on regtest with only one bid? Because in that scenario the names value would be 0.

In [this atomic swap tx for the name `pinkieswear`](https://hnsnetwork.com/txs/32227eb2dab16dbb3debc873ff40cfeba299eb2cf1554ffd098fa968bc88c6cd) you can see the FINALIZE value is 4 HNS.

Input 0 to that transaction spends [this TRANSFER output](https://hnsnetwork.com/txs/294ba9d77ee9c09d5f385e33772370608a0884d46f143d85017ce62bb2678ae4) which also has a value of 4 HNS.

rithvikvibhu: Okay tried it with multiple bids and yeah the input has value (but only when path is also available, 0 otherwise).
So will have to find out how to get covenant info for inputs also to identify the one with names.

rithvikvibhu:
Rebased on master (well actually started from scratch). Now it works properly, tested with names with actual value. Finalize with/without work. Also updated transaction labels.

**Seller's side (no payment, with payment):**
image

**Buyer's side (no payment, with payment):**
image

app/ducks/walletActions.js
rithvikvibhu: As per chat on telegram, updated and doesn't depend on name value anymore.

Issue #203, opened 2020-08-14
Set lookahead in Settings


sdtsui:
User should be able to adjust lookahead so their wallet can support more names/transactions.  Powerusers are asking for this as it looks as if stuff is disappearing from their wallets.

I think this should be a number input.  You might need to restart the node or at least kick off a rescan -> we could make this automatic or have another button for that.

pinheadmz:
Disagree. Look ahead should be a standard. If users set it too low not knowing what it is, they shoot thenself in the foot. It also means they must memorize that number or keep it with their wallet seed if they ever import. I think it should not be user configurable. What we really need are deep wallet refractors in hsd. 

More: https://github.com/handshake-org/hsd/issues/407#issuecomment-668625559

pinheadmz:
Disagree. Look ahead should be a standard. If users set it too low not knowing what it is, they shoot thenself in the foot. It also means they must memorize that number or keep it with their wallet seed if they ever import. I think it should not be user configurable. What we really need are deep wallet refractors in hsd. 

More: https://github.com/handshake-org/hsd/issues/407#issuecomment-668625559

pinheadmz:
Disagree. Look ahead should be a standard. If users set it too low not knowing what it is, they shoot thenself in the foot. It also means they must memorize that number or keep it with their wallet seed if they ever import. I think it should not be user configurable. What we really need are deep wallet refractors in hsd. 

More: https://github.com/handshake-org/hsd/issues/407#issuecomment-668625559

chikeichan:
did some digging into the rescan behavior and agree with @pinheadmz - a high enough fixed value with a rescan button should fix the problem, or at least give users to keep rescanning until they grabbed all their txs.

chikeichan:
The issue I found with lookahead is that if the values is too high, it actually starts to behave inconsistently. I am guessing that the issue might have to do with size limit of bloom filter, if there is one.

The behavior i am noticing is that:
- I have a wallet with a lookahead value of 700
- if i set lookahead value to 500, i need to rescan twice
- if i set lookahead value to 700, i only need to rescan once
- if i set lookahead value to 800, i actually need to rescan twice as well 
- and if i set lookahead value beyond 1000, i start running into errors from wdb

pinheadmz:
These two make sense:

> if i set lookahead value to 500, i need to rescan twice
> if i set lookahead value to 700, i only need to rescan once

This is a bit surprising:

[...]

PR #301, opened 2021-02-12
wallet: add ability to set account depth on first import; remove hsd soft fork

chikeichan:
Main goal of the PR is to remove [kyokan/hsd](https://github.com/kyokan/hsd) fork. Since [this issue](https://github.com/handshake-org/hsd/pull/519) in hsd is not resolved yet, I have also added a new UI feature to manually set account depth on account import. For users who were previously affected by an issue where wallets with high lookahead values could not rescan properly, this provides a way to bypass the issue.

This should allow Bob to go back to using official hsd and resolve a known issue re: rescan loop!

Screen Shot 2021-02-13 at 12 51 49 AM

TODO:
- don't allow account depth to go below 200 (it's already at 200 since it's the default hsd lookahead value)


app/pages/Onboarding/ImportSeedFlow/index.js
```diff
@@ -14,6 +14,7 @@ import OptInAnalytics from '../OptInAnalytics';
 import { clientStub as aClientStub } from '../../../background/analytics/client';
 import { showError } from '../../../ducks/notifications';
 import SetName from '../SetName';
+import UpdateAccountDepth from "../UpdateAccountDepth";
```
sdtsui, line 17: at component level the concept of depth (sing.) seems appropriate.  Why not reflect that in the UI?

just thinking from a hypothetical user's perspective, "I am going to change my Depth**s**" sounds awkward in my head

app/pages/Onboarding/ImportSeedFlow/index.js
```diff
@@ -23,6 +24,7 @@ const CREATE_PASSWORD = 'CREATE_PASSWORD';
 const ENTRY_STEP = 'ENTRY';
 const OPT_IN_ANALYTICS = 'ANALYTICS';
 const SET_NAME = 'SET_NAME';
+const UPDATE_ACCOUNT_DEPTH = 'UPDATE_ACCOUNT_DEPTH';
```
sdtsui, line 27: yeah we wouldn't say "update_account_depths" here, even though there are two depth values, it makes more sense to me

app/ducks/walletActions.js
```diff
@@ -102,9 +112,15 @@ export const fetchWallet = () => async (dispatch, getState) => {
       ...getInitialState().balance,
     },
     apiKey,
+    changeDepth,
+    receiveDepth,
   }));
 };
 
+export const setAccountDepths = (changeDepth = 0, receiveDepth = 0) => async () => {
```
sdtsui, line 120: see comment on depth

app/background/wallet/service.js
```diff
@@ -186,6 +185,36 @@ class WalletService {
     return res;
   };
 
+  updateAccountDepths = async (changeDepth, receiveDepth) => {
```
sdtsui, line 188: see comment on depth

app/background/wallet/client.js
```diff
@@ -22,6 +22,7 @@ export const clientStub = ipcRendererInjector => makeClient(ipcRendererInjector,
   'estimateTxFee',
   'estimateMaxSend',
   'removeWalletById',
+  'updateAccountDepths',
```
sdtsui, line 25: see comment on depth

app/background/wallet/service.js
```diff
@@ -846,6 +874,7 @@ const methods = {
   estimateTxFee: service.estimateTxFee,
   estimateMaxSend: service.estimateMaxSend,
   removeWalletById: service.removeWalletById,
+  updateAccountDepths: service.updateAccountDepths,
```
sdtsui, line 877: depth

app/pages/Onboarding/ImportSeedFlow/index.js
```diff
@@ -135,9 +153,14 @@ class ImportSeedFlow extends Component {
   }
 
   finishFlow = async mnemonic => {
+    const {accountDepth} = this.state;
     this.setState({isLoading: true});
     try {
       await walletClient.importSeed(this.state.name, this.state.passphrase, mnemonic);
+      if (accountDepth > 200) {
+        await walletClient.updateAccountDepths(accountDepth, accountDepth);
```
sdtsui, line 161: depth (and 156)

app/pages/Onboarding/UpdateAccountDepth/index.js
```diff
@@ -0,0 +1,68 @@
...
+          totalSteps={totalSteps}
+          onBack={onBack}
+          onCancel={onCancel}
+        />
+        <div className="update-account-depth__content">
+          <div className="update-account-depth__header_text">
+            Set Initial Account Depth
+          </div>
+          <div className="update-account-depth_subheader">
+            We recommend setting your initial account depth to total number of transactions, and round up to the nearest 100.
```
sdtsui, line 46: How about

`New users may safely ignore. A higher initial account depth configures your node to search for more transactions.  We recommend that power users set initial account depth to their wallet's total number of transactions.`

sdtsui (approved):
copy changes. merge when ready

pinheadmz:
If you are going to switch back to handshake-org/hsd v2.3, the wallet lookahead is only one byte and the default value is `200`, essentially already the maximum. Why would a user ever need to reduce this? They won't be able to increase it because of the 8-bit limit. 

This will also require (I'm pretty sure) a walletDB migration for all users. Bob currently [forces a lookahead of 1000](https://github.com/kyokan/hsd/commit/ead939772fb70d7ed75f9237ba71bf756f0f95dd) but that value isn't written to the database, meaning everyones' database still has lookahead `10`. Only brand new wallets will get a new lookahead value, unless there is a migration (maybe we should include a migrate script in hsd?).

The increased lookahead in hsd from 10 to 200 should be sufficient for most users. The most active wallets may require a second or third rescan to fully recover.

I started working on a real fix for this -- re-re-scanning a block if lookahead is exceeded: 

[branch](https://github.com/pinheadmz/hsd/commits/abortrescan1))

but after finally talking to JJ about it and reading [this bcoin issue]( https://github.com/bcoin-org/bcoin/issues/929) I realized my approach was wrong. So I abandoned that and went back to reviewing other PRs and addressing dns issues, etc. I'll try to get back on it starting next week.

As you've discovered it's a big mess. Not to mention that hsd must continue to support full node, spv node, and remote wallet node. The asynchronicity between wallet and node is a big issue. In fact, the current Bob construction with remote wallet node may start to cause bigger issues during rescans since the wallet is totally async from the node, and as blocks get added it can fill up memory: https://github.com/bcoin-org/bcoin/issues/929#issuecomment-577864980

pinheadmz:
We've also already discussed user configurable lookahead: https://github.com/kyokan/bob-wallet/issues/203#issuecomment-674021399

chikeichan:
Thanks for the review @pinheadmz 

> If you are going to switch back to handshake-org/hsd v2.3, the wallet lookahead is only one byte and the default value is 200, essentially already the maximum. Why would a user ever need to reduce this? They won't be able to increase it because of the 8-bit limit.

We did not change data type to `u32` so no migration is require.

> We've also already discussed user configurable lookahead: #203 (comment)

We are not changing lookahead value, we are only pre-generating address keys before first rescan. Lookahead value stay consistent with `hsd` default.

> In fact, the current Bob construction with remote wallet node may start to cause bigger issues during rescans since the wallet is totally async from the node, and as blocks get added it can fill up memory

happy to discuss on a separate issue as it's unrelated to this PR.


------
Really like to be able to merge back to `hsd` 0.2.3 while still handle issues for wallet with large lookahead. I know of at least one user's wallets with a lookahead value of 4k+ currently and it doesn't rescan properly with 0.2.3 without this fix.

pinheadmz:
@chikeichan OHHHhhhh gotcha so sorry for misinterpreting. Yeah I see, you're setting the depth not the lookahead, that is clever. The wallet just derives a ton of keys in advance, concept ACK!

chikeichan:
A couple new UX changes

Screen Shot 2021-02-18 at 2 37 02 PM
Screen Shot 2021-02-18 at 2 36 53 PM

app/background/wallet/service.js
```diff
@@ -185,6 +185,40 @@ class WalletService {
...
+    const b = this.node.wdb.db.batch();
+
+    account.changeDepth = changeDepth;
+    account.receiveDepth = receiveDepth;
+    await account.save(b);
+
+    if (changeDepth) {
+      for (let i = initChange; i < changeDepth; i++) {
+        const key = account.deriveChange(i);
+        await account.saveKey(b, key);
```
pinheadmz, line 208: So I think this is missing something actually: if you look at `account.syncDepth()` it also sets the depth in the account metadata which then gets written to the DB. If I'm reading this correctly, what this means is that although the wallet may have indeed saved all the addresses to the database, it still *thinks* its depth is:

`latest address index + lookahead` 

I'm not sure what the effect of this will be, it might just mean that the wallet will try to derive and save new addresses (that it already knows about, so no problem) up until the user-selected account depth actually matches the wallets internal account depth. In other words, the walletDB may have all the address, but the wallet/account object doesn't "know" that.

When users ask Bob for a new receive address, they will probably get the address at the wallet's internal depth index, which may be an address they have already used.

package.json
```diff
@@ -147,7 +147,7 @@
     "electron-updater": "4.1.2",
     "history": "4.7.2",
     "hs-client": "https://github.com/handshake-org/hs-client.git",
-    "hsd": "https://github.com/kyokan/hsd/tarball/1e1292d",
+    "hsd": "https://github.com/handshake-org/hsd/tarball/a94ce87",
```
pinheadmz, line 150: hsd master branch should be considered unstable, I think it'd be more prudent to follow hsd releases either from github or npm (latest is v2.3.0) I think this is fine for now, we need a better hsd release process anyway with release candidates etc to ensure we don't release a bug. Whats interesting about you using this commit in particular is that now Bob will have the "presigned REVEAL" endpoint API: https://github.com/handshake-org/hsd/pull/529

pinheadmz (commented):
Sorry I am so late to review this, I left a few things to think about and the warning about address reuse may be worth looking into / fixing before release.

app/background/wallet/service.js
pinheadmz: Ok I see that actually in https://github.com/kyokan/bob-wallet/pull/301/commits/646164ff3758fc326f5b006e8ed03639f7da4419 is where the accounts depths are being set. So, this is bad because then users are going to put in a high number like 10,000 and end up giving the 10,000th address away for payment, and then not realize they will always need to generate 10,000 addresses every time they recover the wallet.

PR #308, opened 2021-02-26
wallet: add deepclean and wdb backup

chikeichan:
This PR add `deepclean` and `backup` to wallet setting view. Deep clean fixes many issues around rescan, and should allow Bob users to upgrade without removing wallet directory going forward.

Also removed `updateAccountDepth` from setting.

Screen Shot 2021-02-26 at 1 09 18 PM
Screen Shot 2021-02-26 at 1 09 12 PM


app/pages/Settings/index.js
```diff
@@ -69,6 +76,15 @@ export default class Settings extends Component {
     transactions: PropTypes.object.isRequired,
   };
 
+  constructor(props) {
+    super(props);
+    this.state = {
+      changeDepth: props.changeDepth,
+      receiveDepth: props.receiveDepth,
+      isUpdatingDepth: false,
```
pinheadmz, line 84: Why do we need these in the state again?

app/pages/Settings/index.js
```diff
@@ -195,7 +222,22 @@ export default class Settings extends Component {
...
+          <div>
+            <div>{`Back up wallet database and files to a directory.`}</div>
+          </div>,
+          'Backup WalletDB',
+          this.onBackupWDB,
+        )}
+        {this.renderSection(
+          'Deep Clean and Rescan Wallet',
+          <div>
+            <div>For more information on deep clean, please see <Anchor href="https://github.com/handshake-org/hsd/blob/master/CHANGELOG.md#wallet-api-changes-1">here</Anchor></div>
```
pinheadmz, line 236: This link may change as new versions get added to the hsd changelog -- I'm not sure a good static link to include here, maybe the PR? https://github.com/handshake-org/hsd/pull/511

PR #314, opened 2021-03-21
add register all functionality

chikeichan:


app/background/wallet/create-register-all.js
```diff
@@ -0,0 +1,65 @@
...
+        continue;
+
+      const output = new Output();
+      output.address = coin.address;
+      output.value = ns.value;
+
+      output.covenant.type = types.REGISTER;
+      output.covenant.pushHash(ns.nameHash);
+      output.covenant.pushU32(ns.height);
+      output.covenant.push(EMPTY);
```
rithvikvibhu, line 53: Hey nice, this is really helpful! Not sure if this is the right place, but is it possible to set records here? Maybe with (non-persisting) user input. Personally, I set it to Namebase DNS so it just works after transferring to Namebase.

chikeichan: Yep, this would be the right place to set records here. For UI/UX i would use the multi-select modal from bulk transfer #315

PR #315, opened 2021-03-21
add bulk transfer functionality

chikeichan:


pinheadmz:
Just wanna say... this is the kind of thing that would be a useful PR to hsd, then everyone can use it ;-) and for Bob integration it would just be a simple API call.

PR #326, opened 2021-04-08
domain-manager: fix expiry value in table

### LINKED ISSUE #318: Inaccurate domain expiry time [v0.7.0-rc.2]
I was looking at my recently transferred in domain name and it said there's 2 years (-1 day) left until expiration, yet in the domain overview it says that there is only "~a year" until expiration.

However, another name I have expires a couple months sooner but still accurately says "~2 years" until expiry.

Guessing the bug has to do with it being recently transferred (and expiration clock reset).
Screen Shot 2021-03-27 at 5 59 45 PM
Screen Shot 2021-03-27 at 5 59 49 PM

rithvikvibhu:
Fixes #318.

Uses `ns.renewal` and the network's `renewalWindow` instead of `ns.height`.

image
image
(for ref, taken today - 2020-04-08)

### DIFF
--- a/app/pages/DomainManager/index.js
+++ b/app/pages/DomainManager/index.js
@@ -389,17 +389,18 @@ export default withRouter(
 const DomainRow = connect(
   state => ({
     names: state.myDomains.names,
+    network: state.node.network,
   }),
 )(_DomainRow);
 
 function _DomainRow(props) {
-  const { name, names, onClick } = props;
+  const { name, names, onClick, network } = props;
   return (
     <TableRow key={`${name}`} onClick={onClick}>
       <TableItem>{formatName(name)}</TableItem>
       <TableItem>
         <Blocktime
-          height={names[name].height + 105120}
+          height={names[name].renewal + networks[network].names.renewalWindow}
           format="ll"
           fromNow
         />


app/pages/DomainManager/index.js
```diff
@@ -389,17 +389,18 @@ export default withRouter(
...
 function _DomainRow(props) {
-  const { name, names, onClick } = props;
+  const { name, names, onClick, network } = props;
   return (
     <TableRow key={`${name}`} onClick={onClick}>
       <TableItem>{formatName(name)}</TableItem>
       <TableItem>
         <Blocktime
-          height={names[name].height + 105120}
+          height={names[name].renewal + networks[network].names.renewalWindow}
```
pinheadmz, line 403: Looks right to me - untested ACK 👍

pinheadmz:
There's also a big `stats` object in every namestate by the way that has a lot of these kinds of things computed already (this may even help simplify the code around bid / reveal phase etc in other parts of Bob):

See the very bottom of this JSON:

```
$ hsd-rpc getnameinfo bob
{
  "start": {
    "reserved": false,
    "week": 25,
    "start": 27216
  },
  "info": {
    "name": "bob",
    "nameHash": "b5d577dc9ce59725e29886632e69ecdf3b6ca49c0a14f4315a2404fc1508672d",
    "state": "CLOSED",
    "height": 27236,
    "renewal": 29434,
    "owner": {
      "hash": "10a6b35f3ebe86996357f911b319fd4a49894810bec171ed701b1963c0348465",
      "index": 0
    },
    "value": 4096000000,
    "highest": 5000000000,
    "data": "0002036e733103626f62002ce706b701c002",
    "transfer": 0,
    "revoked": 0,
    "claimed": 0,
    "renewals": 0,
    "registered": true,
    "expired": false,
    "weak": false,
    "stats": {
      "renewalPeriodStart": 29434,
      "renewalPeriodEnd": 134554,
      "blocksUntilExpire": 71995,
      "daysUntilExpire": 499.97
    }
  }
}
```

rithvikvibhu:
Nice to have those stats, and I think they even change based on the name's current state.

The table in Domain Manager gets the details with `getnames` and that doesn't have stats. Instead of making `getnameinfo` calls for each of name, this felt simpler/easier.

PR #327, opened 2021-04-09
Add ShakeDex fee support

kurumiimari:
This PR upgrades Bob to support ShakeDex fees. Note that the API endpoints are currently disabled on shakedex web, so all mainnet fees will be zero.

Screenshots are attached. Some janky UI I wasn't sure how to fix, but the bones should be there!

Screen Shot 2021-04-09 at 12 38 25 AM
Screen Shot 2021-04-09 at 12 37 44 AM

### DIFF
--- a/app/background/shakedex/client.js
+++ b/app/background/shakedex/client.js
@@ -15,4 +15,5 @@ export const clientStub = ipcRendererInjector => makeClient(ipcRendererInjector,
   'restoreOneFill',
   'getExchangeAuctions',
   'listAuction',
+  'getFeeInfo',
 ]);
--- a/app/background/shakedex/service.js
+++ b/app/background/shakedex/service.js
@@ -110,6 +110,7 @@ export async function fulfillSwap(auction, bid, passphrase) {
     publicKey: auction.publicKey,
     paymentAddr: auction.paymentAddr,
     price: bid.price,
+    fee: bid.fee,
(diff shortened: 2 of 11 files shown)


chikeichan:
Hey @kurumiimari ! This looks good to me. The only changes i added is to include `feeInfo.address` in the ConfirmFee modal. I also make it so that clicking on the address will open up the address view in block explorers. See below.

Screen Shot 2021-04-18 at 7 18 40 PM

chikeichan:
@kurumiimari  I also remove fees from the table view. Since as the buyer, it seems confusing that I might be paying for **additional** fee. Base on the pre-sign that it generates, the `price` is inclusive of `fee` amount, am I correct?

See simnet pre-sign below:

```
    {
      "fee": 120000,
      "lockTime": 1618745166,
      "price": 12000000,
      "signature": "a52db695d641b0aa4185fd2c61e841673e35f8e6b18ce083cd0b4f515dec387a011bbc5fc2dbe304186f2a6b7d1030ae38e1dbff9b8f149fe047654f710299a184"
    }
```

kurumiimari:
`price` is currently exclusive of `fee` amount - the fee paid is in addition to the amount paid to the buyer. I can modify this in ShakeDex without changing the API if this is confusing.

DIPMR:
I think burning is adding only value to price/speculation nothing else. Better not to burn such a scarce asset for fee. Use it for TLD transactions only.

PR #329, opened 2021-04-11
Your Bids: Stats

rithvikvibhu:
I've personally lost ~6 bids at once forgetting to reveal (kept track of dates without time/timezone :[ ). It's also hard to find out what the difference in Total and Spendable balance is for.

This PR adds a few numbers above the table in the Your Bids page:
- Number of bids to reveal
- The time to reveal at the latest to not lose any HNS
- Number of bids that can be redeemed
- How much you get back by redeeming [todo]

While typing this, just realized that you can also get back some HNS by registering, can probably add that too.

These are the only numbers I could think of that are useful, open to any other interesting values to display.
If there's a lot, can even move it to a new page. But for now, I think the "reveal within X hours" is the most important (for me, probably others) so put it in the Your Bids page.

image

To calculate the values, I didn't want to make calls for each name (because some would have thousands of bids). This uses `createRevealAll` and `createRedeemAll` and reads from the created transaction's inputs (and covenant items, thanks @pinheadmz) and only makes a single call.

If there's any better way to do this, will update it.


app/background/wallet/service.js
```diff
@@ -351,6 +351,20 @@ class WalletService {
     return value - amount;
   };
 
+  createRevealAll = async () => {
+    const {wdb} = this.node;
+    const wallet = await wdb.get(this.name);
+    await this._ensureSynced();
+    return wallet.createRevealAll().catch(() => null);
```
RobertLowe, line 358: `catch  () => null` no error handling or feedback to user?

app/background/wallet/service.js
```diff
@@ -351,6 +351,20 @@ class WalletService {
...
+  };
+
+  createRedeemAll = async () => {
+    const {wdb} = this.node;
+    const wallet = await wdb.get(this.name);
+    await this._ensureSynced();
+    return wallet.createRedeemAll().catch(() => null);
```
RobertLowe, line 365: same as above, no error handling

app/background/wallet/service.js
```diff
@@ -879,6 +893,20 @@ class WalletService {
     });
   }
 
+  _ensureSynced = async () => {
+    const nodeHeight = (await this.nodeService.getInfo()).chain.height;
+    const wallet = await this.node.wdb.get(this.name);
+    return new Promise(resolve => {
+      const intervalHandle = setInterval(async () => {
+        console.log('Waiting for wallet sync:', wallet.wdb.height, '/', nodeHeight)
```
RobertLowe, line 901: Remove console statement?

RobertLowe (changes requested):
Minor changes

rithvikvibhu:
Thanks @RobertLowe for reviewing, #358 replaces this so going to close this PR. In the new one, the `create*` methods aren't used anymore.

PR #333, opened 2021-04-16
shakedex: query finalize tx when not found in db

chikeichan:
This fixes an issue where if user finalize an exchange transfer in my domain instead of exchange, bob will look for the the finalize coin from the full node.

TODO:
- write exchagne finalize to exchange db even when user is finalizing it from "My Domain" view


chikeichan:
Now shakedex service always query finalize tx by transfer hash + transfer name. Less dependent on what's written to the db.

CRYONE7:
Are there any much simpler way to make this downloads and just simply verify if i am eligible or not...Why do have to take all this HARDTIME/BLAMING/BLOCKING ON ONE PERSON?

PR #334, opened 2021-04-19
Change node directory and persist api keys

chikeichan:


app/pages/Settings/APIKeyModal.js
```diff
@@ -0,0 +1,93 @@
...
+              Generate New key
+            </a>
+          </div>
+          <input
+            type="text"
+            className="settings__input"
+            value={this.state.apiKey}
+            onChange={e => this.setState({ apiKey: e.target.value })}
+            onFocus={e => e.target.select()}
+          />
```
rithvikvibhu, line 72: THANKS @chikeichan! Was honestly a pain till now that it kept changing

PR #347, opened 2021-05-10
Airdrop Claim

rithvikvibhu:
This PR adds a wizard to claim rewards from the airdrop by providing SSH/PGP keys. It's also possible to skip it and directly input the proof like before.

The file `app/background/claim/airdrop.js` is taken from the [hs-airdrop repo](https://github.com/handshake-org/hs-airdrop/blob/master/bin/hs-airdrop), modified to export functions instead of as a cli script, and also pass along status updates.

A new `ducks/claims` handles logic for creating the proof, it can probably used for name claims also.

Screenshot of first screen only, not all:
image

### DIFF
--- a/app/background/claim/airdrop.js
+++ b/app/background/claim/airdrop.js
@@ -1,0 +1,655 @@
+/* Adapted from https://github.com/handshake-org/hs-airdrop/blob/master/bin/hs-airdrop */
+
+process.env.NODE_BACKEND = 'js'
+
+const { app } = require('electron')
+
+const assert = require('bsert')
+const fs = require('bfile')
+const request = require('brq')
+const Path = require('path')
+const bio = require('bufio')
+const pgp = require('bcrypto/lib/pgp')
+const ssh = require('bcrypto/lib/ssh')
+const bech32 = require('bcrypto/lib/encoding/bech32')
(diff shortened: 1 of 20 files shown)


Falci:
Bob Wallet is an alternative. Not all users are that concerned with copy/paste keys, or they may have different levels of trust on their device/environment. 
The way I see, it's `hs-airdrop` job to make it as much secure as it could be. Bob's job is to make it more user friendly.

rithvikvibhu:
>I think opening a file would be at least feel a little more secure.

Will add a file select then. Will need to consider both armored and binary for pgp, but shouldn't be a problem.

About the clipboard clear, it's not really a bulletproof solution because there are clipboard managers that store a history - in both Windows (built-in) and Linux (kde at least).
Don't want to provide a false sense of sec and let people rely on that. We could add a text reminding them to clear it manually though.

chikeichan:
Hey @rithvikvibhu ! Nice work 👍 

I have made a PR against your branch [here](https://github.com/rithvikvibhu/bob-wallet/pull/1). Please review and let me know if you have any questions!

rithvikvibhu:
Thanks, merged your changes

PR #357, opened 2021-06-02
Reserved Name Claims

pinheadmz:
This adds reserved name claim functionality to the UI.

~DO NOT MERGE until commit deca1a5f1a6c3076560e71194d58046595d254ff is removed~ ✅  OK

## TODO:
- [x] Add Method 3: sign offline and paste RRSIG
- [x] Add Method 4: sign offline using PKCS11 for TLD owners who manage with OpenDNSSEC + HSM
- [x] Add "Success" modal when Claim is broadcast (note that claims do not appear in the mempool!)
- [x] MAKE IT PRETTY (@rithvikvibhu and @chikeichan have kindly offered to polish the UI for me)

## To test:

This branch uses a hard-fork of hsd to test, which includes two extra reserved names `hns-claim-test-1.xyz` and `hns-claim-test-2.xyz`, which are two real ICANN domains that I own and set up for this purpose. When you launch Bob using `npm run dev` it will automatically start in regtest mode and generate a few blocks, including the block I used in the claim commitment. 

To claim `hns-claim-test-1` you must create a wallet with my favorite seed phrase of all time:

```
abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about
```

...and then you can use "Method 1" which is the easiest, since I have already pasted the claim TXT in "the" DNS:

```
$ dig hns-claim-test-1.xyz txt +short
"hns-regtest:aakkrwicqqs2s5aoxavbcaariycxuze3i5fp3ub6n2r62zx3tjtl3ry56orjwonppjdru3p2uirlbyglvvtcepoqbhdacaaaadjczi34"
```

To claim `hns-claim-test-2` you need my DNSSEC private key on your own computer which you then use to sign with "Method 2". That private key is posted [here](https://gist.github.com/pinheadmz/4712b2db28d14d0071ed30868f0410ef). Note that the filename on your local machine must match this filename exactly: `Khns-claim-test-2.xyz.+008+27259.private`

## Navigating

The reserved name claim page is currently accessible in two ways: By looking up a reserved name, or by clicking "Airdrop / Name Claim" in the sidebar, then "Claim":

Screen Shot 2021-06-02 at 2 00 45 PM
Screen Shot 2021-06-02 at 2 00 30 PM

## Usage

Screen Shot 2021-06-02 at 2 01 54 PM


app/pages/Reserved/index.js
```diff
@@ -0,0 +1,379 @@
...
+            <tr><td>TXT</td><td className="wrap">{this.state.claim.txt}</td><td><CopyButton content={this.state.claim.txt} /></td></tr>
+          </table>
+        </div>
+        <div>
+          <h2>Sign and broadcast this claim:</h2>
+
+          <div className="reserved__method">
+            Method 1 (online): Paste above TXT into legacy DNS.
+            <Tooltipable
+              tooltipContent="TODO: blah blah blah DNSSEC, GoDaddy, then come back here and click...">
```
Falci, line 335: image
🤣

app/pages/Reserved/index.js
```diff
@@ -0,0 +1,379 @@
...
+        <table className="reserved__keyList">
+          <tr><td>Key tag</td><td>Key type</td><td>Key Algorithm</td><td>Filename</td></tr>
+          {rows}
+        </table>
+      </div>
+    );
+  }
+
+  render() {
+    const fakesig =
```
Falci, line 301: What do we need this `fakesig` ?

pinheadmz: oh you got me -- I started working on Method 3 where the user signs completely offline and just pastes their RRSIG into Bob. I had a `textarea` where they paste the sig and this was just for a placeholder. Decided to cut it out for now and push what I had... will pick it up again next week.

app/pages/Reserved/index.js
pinheadmz: yeah... hence "draft" mode :-P

chikeichan: looks alright to me

chikeichan:
just tested this - i will start thinking about how the UX would look. would love to target this for `0.9.0` release. (end of June-ish)

one question - how come the claim TX doesn't show in mempool? is there anyway to tell that its pending?

pinheadmz:
> one question - how come the claim TX doesn't show in mempool? is there anyway to tell that its pending?

So, because name claims and airdrop proofs are merged into the coinbase TX, they are handled separately from regular TXs. In fact if you look at the p2p protocol there are additional sets of messages for sending claims & airdrops separate from regular TXs. When received, they are inserted into special buckets in the mempool, and you can see them tallied separately in `getinfo`:

```
{
  "tx": 71,
  "size": 262376,
  "claims": 0,
  "airdrops": 0,
  "orphans": 1
}
```

SO -- all the "get mempool" type APIs that we have in hsd just ignore those extra two buckets. In short, we just don't have the APIs yet (as far as I know) to return claims & proofs. They are a bit tricky to display in the mempool anyway because the hash you get when you send a claim or proof is NOT a txid and will never show up on the blockchain as such, its just a temporary ID that gets thrown away once the proof or claim is merged into a coinbase transaction.

One thing that is badly needed not just for claims and proofs but for ALL TXs is [`rpc testmempoolaccept`](https://chainquery.com/bitcoin-cli/testmempoolaccept) which will indicate to the user if anything is invalid about their TX before they send it. For hsd, we would modify it to also test claims & proofs, and then we could try to give better feedback to the user.

pinheadmz:
> just tested this - i will start thinking about how the UX would look. would love to target this for `0.9.0` release. (end of June-ish)

I think this is a reasonable goal. I can definitely get method #3 (copy/paste TXT + RRSIG) implemented by then but I'm just starting to understand what is needed for PKCS#11. I think it's cool to include that button in the UI but disabled with no route yet as we work on it, and just add that feature later.

pinheadmz:
Screen Shot 2021-06-18 at 2 48 48 PM


Added success modal, implemented method 3 and 4 -- although 4 is a bit different now. I'm going to move the HSM/PKCS11 stuff to a fork of bns so HSM users don't need to install the entire GUI. So now method 4 is just "paste the entire base64 blob here" and calls sendrawclaim. You don't even need to have a reserved name in the search bar for it to work.

app/pages/Reserved/index.js
pinheadmz: Ok I filled in all the `TODO` tooltips... ;-)

app/pages/Reserved/index.js
pinheadmz: this is now in use for method 3

PR #358, opened 2021-06-02
Portfolio cards and Balance break down

rithvikvibhu:
Supersedes and closes #329.

## Balance
This replaces the single "Locked" balance with 3 numbers, shown only when applicable:
- **Locked in live auctions** = bids in BIDDING state auctions
- **Locked in reveal** = bids in REVEAL state auctions (includes blind if not revealed yet)
- **Locked in finished auctions** = bids that can be redeemed or registered

There's also a USD estimate based on coingecko's public API.

image

## Cards
Below the balance section, a row of cards show actionable info that may or may not be time-constrained:
- **Reveal** - Bids not yet revealed, along with the earliest-reveal-period-ending time
- **Renew** - Names expiring in the next 2 months
- **Redeem** - Redeeming lost bids and recovering HNS
- **Register** - Registering won names and recovering the difference
- Transfers in progress and ready to finalize

Just like locked balance blocks, irrelevant cards are hidden.

Since the information needs to be fetched from different places (and frequently! as it is on the main page), imo it made sense to make one pass over all bids and collect everything instead of calling `createRegisterAll`, `createRedeemAll`, etc. which will go through the same loops multiple times, separately.

image

## Test
Either simply try out random actions (bids, reveals, transfers, etc.) or use this for reference (regest) - after starting at block 200, at block 228, all states are covered (except transfers).
image
Don't have a script to place these bids yet, will try to write one based on https://gist.github.com/pinheadmz/49e3fac7d797a99c3a78fb3ca0ddc012.

### TODO
- [x]  Add code to check "expired" status
- [x]  Function for Finalize All
- [x]  Function for Renew All - clicking on these buttons currently open the Your Bids page
- [x] Remove `console.log()` statements before merging


app/pages/Account/functions.js
```diff
@@ -0,0 +1,256 @@
...
+
+async function getStatsFromNames() {
+  let transferringDomains = new Set();
+  let transferringBlock = null;
+  let finalizableDomains = new Set();
+  let renewableDomains = new Set();
+  let renewableBlock = null;
+
+  const names = await walletClient.getNames();
+  console.log("names", names);
```
RobertLowe, line 163: clean up log statement

app/pages/Account/functions.js
```diff
@@ -0,0 +1,256 @@
...
+      finalizable: { domains: finalizableDomains },
+    },
+  };
+}
+
+async function getUsdConversion() {
+  try {
+    const response = await (
+      await fetch(
+        "https://api.coingecko.com/api/v3/simple/price?ids=handshake&vs_currencies=usd"
```
RobertLowe, line 224: could this be a setting? usd/cad/eur etc

RobertLowe (changes requested):
Also, the 80 + ## + ## breakdown is nice, but also show the = total lockup as well as it currently does.

Design note, not sure we need the colors, but no strong preferences here. Reveal buttons style is not consistent with other buttons in the app

👍

app/pages/Account/functions.js
rithvikvibhu: Hm yeah, can do that :+1:

app/pages/Account/functions.js
rithvikvibhu: Yeah, that's the last item on the todo list above. Keeping it for now as people review, helps find any numerical differences. Will remove before changing this PR from Draft to Ready to Merge.

rithvikvibhu:
Thanks for the feedback @RobertLowe!

>Also, the 80 + ## + ## breakdown is nice, but also show the = total lockup as well as it currently does.

It takes up space, might have to reduce the font size for everything (it's scrollable, but nice to have the whole breakdown in one view.) Will see if it fits.

>Design note, not sure we need the colors, but no strong preferences here. Reveal buttons style is not consistent with other buttons in the app

The colors indicate how serious the operation is (has a time limit, some pending action, just notices of something happening). We could lighten them a bit more.

I agree, black buttons are weird. Have any suggestions for them?

RobertLowe:
Thanks for keeping on top of this @rithvikvibhu 

@rithvikvibhu @chikeichan I think all new features merging in should require tests, my face went a bit pale when I saw there were only 2 spec files in the entire application :S

rithvikvibhu:
We do need tests, but like you said, the are only 2 of them (one of which is completely commented out).
But I think we need a proper format to write tests and should start with the basics/main features first, especially for large projects like these.
And only when there's a (sort of) standard around how to write them, can new features be required to have tests. Otherwise, we'll have a bunch of half baked tests that may need to be rewritten later (or never).

RobertLowe:
@rithvikvibhu maybe in the future, I'm still learning `hsd`'s tests https://github.com/handshake-org/hsd/blob/master/test I need a better foundation there first, but I'm starting to interact with regtest via my own project tests, maybe after that I can contribute more here

app/pages/Account/functions.js
```diff
@@ -0,0 +1,313 @@
...
+  // If nobody revealed their bids, start over.
+  // if (auction.owner.isNull())
+  if (auction.owner.hash === ZERO_HASH && auction.owner.index === 0xffffffff) {
+    return true;
+  }
+
+  return false;
+}
+
+async function getStatsFromBids() {
```
chikeichan, line 41: i think we should move `getStatsFromBids` and `getStatsFromName` to background, and on UI thread we can just call `walletClient.getBidStats` and `walletClient.getNameStats`

app/pages/Account/functions.js
```diff
@@ -0,0 +1,313 @@
...
+}
+
+async function getUsdConversion() {
+  try {
+    const response = await (
+      await fetch(
+        "https://api.coingecko.com/api/v3/simple/price?ids=handshake&vs_currencies=usd"
+      )
+    ).json();
+    return response.handshake.usd ?? 0;
```
chikeichan, line 276: let's use `||` instead of `??`

app/background/wallet/service.js
```diff
@@ -446,6 +447,12 @@ class WalletService {
     await finalizeMany(wallet, names);
   };
 
+  renewMany = async (names) => {
```
chikeichan, line 450: not blocking for this PR - now that we have ledger, we should add `ledgerProxy` to all calls like this. We will need to separate `renewMany` to `createRenewMany` and `sendRenewMany` for `ledgerProxy`.

app/pages/Account/functions.js
```diff
@@ -0,0 +1,313 @@
...
+        }
+      }
+      continue;
+    }
+
+    // At this point, the bid is revealed (at least)
+
+    // If still in reveal period, the blind is no longer locked
+    if (auction.state === "REVEAL") {
+      revealingHNS -= blind;
```
chikeichan, line 134: Screen Shot 2021-06-11 at 9 23 01 PM

I found an edge case where there are `revealingHNS` but no `revealableHNS`.  I think you can likely combine the two variables here.

app/pages/Account/functions.js
```diff
@@ -0,0 +1,313 @@
...
+  let transferringBlock = null;
+  let finalizableDomains = new Set();
+  let renewableDomains = new Set();
+  let renewableBlock = null;
+
+  const names = await walletClient.getNames();
+  console.log("names", names);
+
+  for (let name of names) {
+    const auction = await walletClient.getAuctionInfo(name.name);
```
chikeichan, line 204: I would call the following directly instead of `walletClient.getAuctionInfo`, since each rpc call is an http request that has overhead. If a wallet has thousands of name, it might be a bit slow.

```
const ns = await wallet.getNameStateByName(name);
json = ns?.getJSON(this.lastKnownChainHeight, this.network);
```

RobertLowe:
Did the UI change? 

Screen Shot 2021-06-13 at 3 59 03 AM


I liked the breakdown in 
ui

RobertLowe (commented):
Did the UI change? 

Screen Shot 2021-06-13 at 3 59 03 AM


I liked the breakdown in 
ui

rithvikvibhu:
The main changes:
Before:
- The first locked value was only for bids in BIDDING state
- The second locked value was for bids in REVEAL state (including true bid values and not-yet-revealed blinds)

Now:
- The first locked value is for bids in BIDDING as well as REVEAL (the true bid values that cannot be recovered for ~10 days)
- The second locked value is for bids in REVEAL and not-yet-revealed (does not include true bid values, this is moved to first value)

This was done mainly because the "In reveal" label may be confusing as only a part of it would be immediately recoverable.
There are still 3 parts for Locked Balance, I think it just happens that (in your pic) all locked values fall in the first category so only 1 value is shown.
The only other UI change is removing the card buttons and making the whole card clickable.

RobertLowe:
I don't see the Real 1 Bid / Renew / Redeem 1 bid, etc

Would this be because I just don't have any?

rithvikvibhu:
Yup, only actionable cards (with an exception of transfers, which stay until finalized) are shown.

RobertLowe (approved):

chikeichan:
@RobertLowe thank you for reviewing our PR! please join our discord if you are interested for more discussion on bob - https://discord.gg/qtpmPCgrCB

PR #363, opened 2021-06-11
yourBids: add filter by bid status

chikeichan:
`getBids` will re-index every bids calling `wallet.getNameStateByName`. This limits network/rpc even locally, as even local rpc call is still a web request that adds overhead. I perf tested this with a wallet that has about 9k bid history on almost as many names, and the loading time with filter feels the same as loading time without filters.

bidfilter

### DIFF
--- a/app/pages/YourBids/index.js
+++ b/app/pages/YourBids/index.js
@@ -19,6 +19,7 @@ import c from "classnames";
 import * as nameActions from "../../ducks/names";
 import * as notifActions from "../../ducks/notifications";
 import dbClient from "../../utils/dbClient";
+import {NAME_STATES} from "../../ducks/names";
 
 const analytics = aClientStub(() => require('electron').ipcRenderer);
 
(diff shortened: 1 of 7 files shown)


app/pages/YourBids/index.js
```diff
@@ -43,18 +45,22 @@ class YourBids extends Component {
...
   };
 
   async componentDidMount() {
     analytics.screenView('Your Bids');
-    this.props.getYourBids();
+    this.props.getYourBids()
+      .then(() => this.setState({ loading: false }));
     const itemsPerPage = await dbClient.get(YOUR_BIDS_ITEMS_PER_PAGE_KEY);
-
+    console.log(this.props.match.params.filterType)
```
RobertLowe, line 60: remove console.log statement

RobertLowe (changes requested):
I was unable to test this, causes the app not to load. When running the `npm package` and `bob.app`.

I can run other branches without issue

rithvikvibhu:
@RobertLowe Are there any errors when running `bob.app` in a terminal? Also does `npm run dev` work?

RobertLowe:
@rithvikvibhu nothing, it just hangs

Screen Shot 2021-06-13 at 4 53 24 AM

RobertLowe:
dev loads

Screen Shot 2021-06-13 at 4 55 55 AM

chikeichan:
i was able to build successfully with `node@12.13.1` and `npm@6.12.1`. You might have to nuke node_modules and reinstall again.

Issue #131, opened 2020-04-23
Allow user to disable idle timeout and display chain stats on lock screen


brandondees:
Since it can take a good long while for synchronization to complete during initial setup, it would be really helpful UX wise to be able to get some idea of how things are progressing, and to help distinguish broken/stuck from just taking a while to finish. I presume it's feasible to fetch the network's current block height from hsd before it's fully caught up, and you could track sync progress relative to that, if I understand correctly.

Furthermore, it'd be nice to see synchronization status while the wallet is locked, since that seems to happen pretty quickly and isn't end-user configurable. I don't need to go to the trouble of fetching the passphrase from my vault each time if I'm just trying to check on whether sync is done yet.

pinheadmz:
Are you talking about initial block download? There is a % sync indicator in the top right corner of the UI.

As far as Bob knowing what the sync progress is, the progress comes directly from hsd and actually given how blockchains work there is no way to really know if you are synced, but we can guess. More details about this here: https://github.com/bcoin-org/bcoin/issues/886

I suppose we could display some non-private details on top of the lock screen like node status and chain height.

brandondees:
After setting up a new wallet just now I'm seeing the top right indicator now on the latest master. Previously (a few days ago) I couldn't get to that point, as the initial pre-wallet screen seemed to take forever with no indication of progress. 

Yeah, the lockout is a bit annoying and it'd be helpful to show some general status info from that view without having to go through the unlock procedure just to check on it. I'd also like to have the option of configuring that lock timeout or disabling it on my trusted system.

pinheadmz:
There is some work on passphrase stuff about to be merged in for the next version:

https://github.com/kyokan/bob-wallet/pull/128

This PR is written to solve the "No Passphrase" error that comes up after a minute or so when you try to send a TX. It separates the actual wallet lock from the UI "logout".

A little background, when you enter your passphrase into hsd's wallet, it decrypts your master private key and keeps it in memory for (default) 60 seconds, then wipes it out again. This is enough time to sign a TX with the private key. My experience with Bitcoin wallets is that you always have to enter your passphrase when sending a TX (for Bob that also means opening an auction, bidding, revealing, updating, etc...) and so this PR enforces that UX.

So after #128 is merged we can address another question: Should Bob UI still lock after 5 minutes of idle time and require the passphrase to reopen the UI? In my opinion, this is a good privacy measure because your balance and transaction activity is visible on the screen.

[...]

PR #372, opened 2021-06-18
Setting for idle timeout

rithvikvibhu:
PR for first half of #131.

This adds a new row in Settings to adjust the inactivity timeout (in minutes) or disable it completely.
It still defaults to 5 minutes as before.

image
image


app/ducks/walletReducer.js
```diff
@@ -40,6 +41,7 @@ export function getInitialState() {
     receiveDepth: 0,
     transactions: new Map(),
     idle: 0,
+    maxIdle: 5,
```
chikeichan, line 44: i think this state should live in `app` duck instead of the `wallet` duck. wdyt?

rithvikvibhu: I placed it in wallet because the state `idle` and actions, reducers `resetIdle`, `incrementIdle` are all part of wallet already. But yeah it's more of a global thing, not wallet-specific. I'm fine with it either way.

PR #373, opened 2021-06-18
Show sync status on lock screen

rithvikvibhu:
PR for Second half of #131.

1. Moves the sync status code to a new component
2. Adds that to AppHeader (lock screen header)

Also cleans up code a bit + lint.

image


chikeichan (approved):
LGTM!

PR #374, opened 2021-06-18
Disable DNS (ns/rs)

### LINKED ISSUE #140: Conflict when hnsd is already running
Running  6b15a2cf61963eef134baa4664df4ec848ce241f built for OSX.

If `hsd` is already running on mainnet when you launch Bob, you get the familiar, easy-to-understand error: 

Screen Shot 2020-05-18 at 10 16 11 AM

However, I've been working on `hnsd` and had it running as my computer's resolver on mainnet (running as root with `--rs-host=127.0.0.1:53`). When I launched Bob this morning I got a much more terrifying prompt:

Screen Shot 2020-05-18 at 10 17 23 AM

Shutting down Bob, shutting down hnsd, then re-launching Bob worked around the issue. My wallet, watchlist, all data, etc -  everything is fine.

rithvikvibhu:
Resolves #140

As per https://github.com/kyokan/bob-wallet/issues/140#issuecomment-630318788, looks like it was decided that wallets and dns resolvers should be kept separate, and I agree.

Having said this, I do feel like running a separate full node to test dns (and have a complete chain copy) is kinda redundant. How about a settings option (default: off) for dns in Bob? @chikeichan @pinheadmz thoughts? If yes, I'll update this PR to add that option.

Now that https://github.com/handshake-org/hsd/pull/554 has been merged in (and Bob has updated to include this), it's as simple as setting `noDns: true` (what this PR is doing right now).

It also removes 2 config options that aren't being used (`listen` and `bip37`) as `publicHost` [defaults to](https://github.com/handshake-org/hsd/blob/2d1cbe9c17b0ad4e8858c06a8f85625dbee35ba9/lib/net/pool.js#L4372) `0.0.0.0` which isn't reachable from "the outside".


chikeichan:
i am in favor of adding a setting that's default to `noDns: true`

rithvikvibhu:
Updated to add to Settings:

(sorry, bad recording)
Peek 2021-06-18 20-03

pinheadmz:
This is restarting the node after changing the setting right? I think a setting is ok - but its really only useful for testing. No one should be keeping Bob Wallet open 24/7 on their laptop to serve DNS. You wouldn't need a separate full node if you're just using hnsd for DNS for example. Or whats more likely is users run Bob in pruned mode or SPV mode and have a separate full, pruned or SPV node running on a server or RPi for DNS service.

rithvikvibhu:
Yeah it's restarting the node.
Agreed, I don't mean as dns resolvers for daily use. More for dev (at least for me) where I don't want to run hsd separately just to test dns - I'd rather start Bob which also lets you manage names. Otherwise will end up with 12 GB (as of now) * 3 and growing (until pruned mode is added).

PR #375, opened 2021-06-20
Ledger: handle *All, *Many transactions

rithvikvibhu:
This PR adds Ledger support for:
- Bulk Transfer (`createTransferMany`)
- Finalize All (`createFinalizeMany`)
- Register All (`createRegisterAll`)
- Renew All (`createRenewMany`)

Afaik, these were the only pending methods for Ledger. If there's any missed, I can include them too.


rithvikvibhu:
The build fail is because of missing `sudo apt update` in CI, will work after #371 is merged (this includes the fix)

pinheadmz:
Awesome dude! I tested with ledger nano S physical device app version v1.0.4 and successfully revealed TEN bids in one TX! Hell yeah.

One minor UI question - when we reveal multiple bids, maybe there is another thing to display here instead of just one of the names (like either all if possible or none?)

Screen Shot 2021-06-21 at 10 51 17 AM

pinheadmz:
Ok, confirmed success with batch register and renew now also...

pinheadmz:
Last test I did was to create a register that required a change output, which was handled perfectly with `_ledgerProxy()` this is all good!

pinheadmz (approved):

PR #384, opened 2021-07-10
Hide zero locked balances

rithvikvibhu:
Hides locked balance items when there are bids, but the HNS value is 0.

**Before:**
image

**After:**
The 2nd item is hidden.


jpjpjo:
How to import BTC

rithvikvibhu:
@jpjpjo Bob Wallet is just a wallet, not an exchange. For a list of exchanges with HNS: https://en.handypedia.org/wiki/Help:FAQ

PR #397, opened 2021-07-28
Background wallet/node refactor: wallet-as-plugin, recover failed connections, clean up and remove unused or redundant code

pinheadmz:
## *DON'T PANIC*

Sorry this is a big one.

The original goal of this PR was to restore the wallet-as-a-plugin configuration when the hsd full node is local (as opposed to remote RPC mode) which is the default. This ended up leading to a ton of refactoring in the node/wallet background services as well as node and wallet actions. There are so many pitfalls and footguns left open for the user that I wanted to cover. Along the way I noticed a good deal of redundant code and unused code and unnecessary code and I cleaned that all up as well. Reviewers may want to read each commit separately (I did a few git rebases so the PR should makes sense like this).

New features:

- By default, hsd node is run locally with wallet as plugin
- User can switch connection to remote full node, in which case wallet is run as `WalletNode` (independent process)

Things we want to be able to do without crashing and without node or wallet state appearing wrong in the UI:

- In any configuration, reload the UI (ctrl-R or change a UI file in dev mode)
- In any configuration, shut down and restart the application
- When switching to remote RPC mode, test the connection FIRST before switching (this includes "correct network" test)
- In any configuration, switch networks. If remote node is on wrong network, allow user to recover by switching to internal node or re-setting remote node options
- If remote node drops the connection, warn user immediately and allow them to recover by reconfiguring

Additional notes to Bob devs:

- The "master" network is now set by the WALLET. The value in the network switcher is set by the wallet, because we might not know anything about the remote full node, so we need to think of Bob from the perspective of the WALLET -- that is attached (one way or another) to some full node, which is secondary.

- UI reloads will wipe out the redux store but not interrupt the background services. That means anything that is "dispatched to main window" from the background must be sent *again* if the UI is refreshed. You can see in the service files in the "start" functions that if the service is already running, I just dispatch that service's relevant properties and return.

- Node details are dispatched by the node service, wallet details are dispatched by the wallet service. I separated the concerns.

- I removed a bunch of lines like "ensure client" when a client isn't actually used in a function. I hope I didn't pull out anything critical but it seemed like there were a lot of unnecessary checks probably left over from even earlier architectures of Bob.

- Both wallet and node services use http clients as well as direct library function calls, I think a good long-term goal would be to get rid of any unnecessary local loopback HTTP calls but a lot of time they are necessary because the hsd RPC modules do extra processing that we'd just have to duplicate in Bob.

- After switching internal/external connections, a user may need to rescan the wallet (especially if the internal node is out of sync) or simply refresh the UI.

- The remote RPC settings are global but could be saved in the Bob db for each network separately

- `isRunning` has a more general meaning now: Basically in either configuration, the START event switches this value to `true` and the STOP event switches it to false. In essence, it refers to the wallet instead of the node. Since an invalid or broken connection sends a STOP event, `isRunning: false` should always mean "no connection" -- remote node is not connected and internal node is not running.

## Demo

https://youtu.be/_87mXSjbxrw


pinheadmz:
@chikeichan I'm happy to just rebase on master branch any time as well, might be cleaner than these github merges like https://github.com/pinheadmz/bob-wallet/commit/7d7f8ca6979f06a84472e10428d46e6b63e6c436 but whatever works for you

chikeichan:
i normally squash and merge anyways, just didn't want to ping you for a simple rebase.

...but i did miss a trailing comma in `app/background/node/client.js:29` and `app/background/wallet/client.js:62` in my merge commit tho, if you could push that changes up before i merge 😓

pinheadmz:
@chikeichan gotcha, fixed.

nlydv:
I've been dealing with some persistent problems with Bob lately where certain actions would result in a blank white UI that would force me to restart the app. Clicking `domain manager` in the sidebar consistently led to that (making the manager unusable). This would also usually occur as soon as I'd log into an alternative wallet/account (i.e. not the first option in the dropdown list).

All I was able to determine was that the bug would _only_ occur when connected to a remote RPC node.
Building this PR on macOS, seems to have (as far as I can tell) solved whatever was leading to those issues.

So from an end-user perspective, I'd give it a positive review.

PR #404, opened 2021-08-23
Fix unreplaced CSP policy value due to sed command in package.sh

nlydv:
In the pre-built macOS release, as well as when built from source, the console shows an "invalid CSP value: {{cspValue}}" error. I saw that the `package.sh` build script tries to use an incompatible GNU version of `sed`'s in-place replacement option.

To make it interoperable for both BSD and GNU sed versions, I changed it to make a .tmp backup of the original file and then delete it right after.

Which defeats the purpose of the "in-place" option flag, but at least it works.


scripts/package.sh
```diff
@@ -2,7 +2,8 @@
 mkdir -p ./dist
 cp app/app.html dist
 CSP_POLICY="default-src 'self'; style-src 'self' 'sha256-GhG3bE0iJoXJDtzwjDYe4ewzpUCrcbsJVwiqGhTOAVg=' https:\/\/fonts.googleapis.com; font-src https:\/\/fonts.gstatic.com; img-src * data:; connect-src http:\/\/localhost:13037 http:\/\/localhost:13039 http:\/\/localhost:14037 http:\/\/localhost:14039 http:\/\/localhost:15037 http:\/\/localhost:15039 http:\/\/localhost:12037 http:\/\/localhost:12039 https:\/\/*.sentry.io https:\/\/*.mixpanel.com;"
-sed -i "s/{{cspValue}}/${CSP_POLICY}/g" dist/app.html
+sed -i.tmp "s/{{cspValue}}/${CSP_POLICY}/g" dist/app.html
+rm dist/app.html.tmp
```
pinheadmz, line 6:
```suggestion
sed -i '' "s/{{cspValue}}/${CSP_POLICY}/g" dist/app.html
```
This works for me cleanly on OSX, is it compatible on windows on linux?
I forgot OSX uses a different `sed`, how annoying. h/t: https://stackoverflow.com/questions/7573368/in-place-edits-with-sed-on-os-x

I think @mslipper originally used `gsed` here (I think he started Bob development on OSX) but @rithvikvibhu changed it to `sed`: 17c08db37e2d7158f4e79bcf8da2a209b73e7715

rithvikvibhu: Windows uses a js script, this is sh script is only for linux/macos. Yeah I changed from gsed to sed here in #350, didn't know sed was "one of those tools". I think the tmp is fine as long as it works.

nlydv: GNU sed requires zero space separating the `-i` and `[opt-val]` so it doesn't work with linux... 🙄
Testing that with sed on linux (same result using gsed on Mac):
```
❯ echo '{{replaceme}}' > test.txt
❯ sed -i '' "s/{{replaceme}}/success/g" test.txt
sed: can't read s/{{replaceme}}/success/g: No such file or directory
```

rithvikvibhu:
Tested, works, app.html is updated, lgtm. Thanks @nlydv!

PR #408, opened 2021-09-09
perf: throttle expensive calls during sync

chikeichan:


pinheadmz:
ACK! Had something similar here: https://github.com/kyokan/bob-wallet/pull/331

PR #417, opened 2021-10-21
fix stuck at confirm seed when creating wallet

chikeichan:


Silkies20:
What is the resolution to this,im having this exxact problem it will not allow you to click to continue to verify the seed

PR #430, opened 2021-11-23
fix collapsed whitespace when rendering record

chikeichan:
Screen Shot 2021-11-23 at 11 18 53 PM

### DIFF
--- a/app/components/Records/Record.js
+++ b/app/components/Records/Record.js
@@ -17,7 +17,7 @@ class Record extends Component {
     const currentTypeIndex = DROPDOWN_TYPES.findIndex((d) => d.label === type);
 
     return (
-      <TableRow className={classNames(this.props.className)}>
+      <TableRow className={this.props.className}>
         <TableItem className="record__type">{DROPDOWN_TYPES[currentTypeIndex].label}</TableItem>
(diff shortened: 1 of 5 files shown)


app/components/Records/Record.js
```diff
@@ -15,9 +17,9 @@ class Record extends Component {
     const currentTypeIndex = DROPDOWN_TYPES.findIndex((d) => d.label === type);
 
     return (
-      <TableRow>
-        <TableItem>{DROPDOWN_TYPES[currentTypeIndex].label}</TableItem>
-        <TableItem>{value}</TableItem>
+      <TableRow className={classNames(this.props.className)}>
```
Falci, line 20:
```suggestion
      <TableRow className={this.props.className}>
```

chikeichan: `classNames()` actually serves a purpose here - it would default `undefined` to `''`

Falci: Oh, I see. Passing `className={undefined}` should be fine because it's optional.

But `TableRow` should use the `classnames` and threat the property as optional
https://github.com/kyokan/bob-wallet/blob/5f6f352a58e320ee4b4bfa8ca62979247958455f/app/components/Table/index.js#L58

Issue #399, opened 2021-08-02
Remove or raise atomic swap limit


peepohns:
The atomic swap feature is [capped at 2000 $HNS ](https://github.com/kyokan/bob-wallet/blob/9ac7b6da645509ff5ef78c9c3d3deea6ccbf10cd/app/background/wallet/service.js#L636-L638). 

I wanted to know if this can be raised or even removed. It's such an awesome feature and one I could see being very very popular especially among sellers switching to a more decentralized marketplace.

rithvikvibhu:
Resolved in #441

PR #441, opened 2022-01-12
Remove atomic swap limit

altslash:
Limit of 2000 HNS for finalize with payment significantly reduces the usefulness of the feature, this PR removes the limit.


pinheadmz:
untested ACK. Closes https://github.com/kyokan/bob-wallet/issues/399

I think it's been around long enough we can remove the limit. Could still be an unknown security issue though. But so far no one to my knowledge has totally lost their payment.

PR #446, opened 2022-01-22
Added Catalan language, Spanish (Spain) language and XXX.json file.

faltrum:
The internationalization of Bob Wallet in version 0.9.0 has been a very important step with the i18n. A new stage is opened so that members of the community can contribute their grain of sand without being professional developers, simply by translating Bob-Wallet into their mother tongue. However, it will be a challenge to put order and do the translation systematically so as not to repeat processes and waste time.

XXX.json file
For this reason, the creation of the XXX.json file is very important, since it is very easy to check which phrases, words have not yet been included in the en.json file. The following steps are proposed:
1.- Load the XXX.json file in Bob-Wallet to detect what is missing to add to the en.json
2.- New lines to be translated in en.json are added. And a specific period of days is opened to carry out the translation and upload the files by the translators.
3.-Then these new lines are also added to the XXX.json file, in this way the translators will be able to detect which lines have been added quickly, without having to check one by one.

On the other hand, I propose the creation of a file called Terminology (located in the LOCALES folder), to specify the meanings of certain technical words in order to translate them correctly into other languages.
Thank you.

PR #447, opened 2022-01-22
Update i18n.js

faltrum:
Updating Spanish and CAtalan LAnguages


rithvikvibhu:
Included as part of #446 and merged in. `ca` and `es-ES` is available in settings.

PR #456, opened 2022-02-01
Sync fixes

### LINKED ISSUE #429: Issue with BOB desktop 9.0 Release Candidate
There are 2 bids to be revealed and it's showing in the "PORTFOLIO" tab, the card, but the reveal bids are not showing in the "BIDS" tab. Even it s not showing the numbers for "bidding()" nor for closed  too. Its showing " but not the breakdown of all bids. ALL" I can reveal the bids from the card though.
It's SPV enabled.
Please find attached screenshots.
1bb
2b

### LINKED ISSUE #448: Bid History: Filters in blank
## Steps to Reproduce

1. Bid
1. Reveal
1. Add To Shakedex a name
1. Open auction

## Context (Environment)

I can't filter by Reveal, Bidding o Closed, because all appear in blank. And when I finalize the acution in Shakedex and recover my domain, it appears like Sold, but It was not sold.
## Screencap
image

### LINKED ISSUE #455: Sync never stops even if wallet catches up to node
`wallet.walletSync` is always true even if both heights are equal.

Even if 100% synced on mainnet, Current height says:
image

and status in header is:
image

**Note:** Rescan was never called, but it still says rescanning.

rithvikvibhu:
fixes #448, fixes #455, fixes #429

`lastKnownChainHeight` was never being set in `Wallet` service. And wallet sync never stopped because the node's height wasn't available (was always 0).
Without lastKnownChainHeight, the namestate of each name used in Your Bids page was being retrieved as of height 0, not current height.

Now, both are updated in Wallet service on events from wallet and node.

### DIFF
--- a/app/background/node/service.js
+++ b/app/background/node/service.js
@@ -355,12 +355,12 @@ export class NodeService extends EventEmitter {
 
     try {
       const info = await this.getInfo();
+      this.emit('refreshNodeInfo', info);
 
       dispatchToMainWindow({
         type: SET_NODE_INFO,
         payload: info.chain,
       });
-
 
       if (info.chain.progress > 0.99) {
         const fees = await this.getFees();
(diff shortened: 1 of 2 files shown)


rithvikvibhu:
Been working for days and no issues, syncs properly.

Issue #438, opened 2022-01-05
Allow "primary" wallet recovery


pinheadmz:
A lot of users accidentally use the `primary` wallet via CLI without realizing it, and it becomes very difficult to use in Bob. I propose, simply:

When user creates new wallet, if the name they provide is "primary", just prompt for a passphrase like usual and encrypt the existing `primary` wallet, don't throw an error and don't create a new wallet. Proceed with the default primary wallet from hsd as new "primary" wallet in Bob

Issue #182, opened 2020-07-28
When hsd is running already, Bob starts up with the "ports in use error" -- I wonder if this error message can also have a link to "switch to your own node"? Same with setting up a node for the first time -- it goes through the entire 24-word seed flow before you have the option to switch to your own node.


chikeichan:
When hsd is running already, Bob starts up with the "ports in use error" -- I wonder if this error message can also have a link to "switch to your own node"? Same with setting up a node for the first time -- it goes through the entire 24-word seed flow before you have the option to switch to your own node.

_Originally posted by @pinheadmz in https://github.com/kyokan/bob-wallet/pull/171#issuecomment-662467434_

pi0neerpat:
Related to this - If Bob cannot connect to your node on startup, it will never go past the "Loading node..." screen. So you will be unable to change your node configs.

Maybe should be a link here as well to "switch to your own node"

![image](https://user-images.githubusercontent.com/35622595/92813263-e1f42180-f38e-11ea-881b-760daf3ef49c.png)

pi0neerpat:
Related to https://github.com/kyokan/bob-wallet/issues/131#issuecomment-621491868

PR #458, opened 2022-02-01
Misc fixes

### LINKED ISSUE #205: [rc-0.5.0] Balance/Pending Tx Issues
Context: Low-usage wallet, <10 names in it.

I notice that when someone sends me funds, it doesn't catch new transactions automatically.  Leaving my computer open when connected to RPC basically seems to do nothing -- I can see stuff confirm on HNScan, but I only see the tx get seen **after** a refresh (ctrl+r) or full restart of the app.

In addition, after the reset/refresh, I notice that the Spendable Balance number doesn't update -- like I can see the inbound funds confirmed in Transaction History, but SB doesn't change until I do yet another refresh.

In summary, I've never seen a pending tx...it seems to be skipping that and only showing received funds.  The expected behavior would be automatically catching Pending txs and UI consistency when updating balances.

Note that this is just for sending funds. I haven't paid as much attention today to what it does when sending bids.

### LINKED ISSUE #378: Usability issue makes you think you won TLD early
Currently if TLD auction is in "Reveal" state and you are current highest bidder than the TLD is listed under "Domains Manager" tab of your Bob Wallet. It even says "~2 years" expiry and HNS Paid balance. The only thing that signifies that you don't own the domain yet is if you try to edit records you'll see some text like "NEXT TREE UPDATE: BLOCK 74700 (~5 HOURS)".

## Expected Behavior
Add a column in the "Domans Manager" with the status of domain auction: "In Reveal" or "Won"

## Current Behavior
It looks like you won.

## Steps to Reproduce
Big on auction and wait until "In Reveal" then find the domain in your "Domain Manager" list in Bob Wallet.

## Context (Environment)
Made me think I won all the domains. Context: :(

Screen Shot 2021-07-02 at 6 16 47 PM

### LINKED ISSUE #394: OPEN doesn't leave pending state in Auction UI
The Auction sidebar says **Your open auction request is pending and will be mined within the next block.** throughout the open period, even after the open is confirmed (and until the bidding period).

image

In latest v0.8.0.

### LINKED ISSUE #396: Highest Lockup NaN HNS error when bidding ends
## Expected Behavior

I should see the highest lockup in number form.

## Current Behavior

`NaN` shouldn't be displayed.

## Steps to Reproduce

Not sure if the decimal has something to do with this error.

## Screencap

SS 2021-07-26 at 11 22 08 AM

### LINKED ISSUE #410: Wallet name 'primary' should show error
Seen on desktop wallet, v0.8, Windows 10

Using this wallet name should not be allowed as it is reserved. Without the error it is possible to proceed to last step where no further buttons are visible.

Expected behaviour: error message should show on first screen.

### LINKED ISSUE #425: Add Funds (aka: Handshake Worldwide Airdrop) shows incorrect reward amount
The 'Add Funds' (aka: Handshake Worldwide Airdrop) section of the app incorrectly states that you can claim `4,662.598321 HNS`, yet the official page states that it is only `4,264.994314 HNS`, and the amount I received appeared to only be `4246.894314 HNS`.

This leads to confusion for the end user, wondering why they didn't seem to receive the full amount.

## Expected Behavior
The app would correctly state that only `4,264.994314 HNS` (assuming the official docs are correct) or `4246.894314 HNS` (assuming the amount I received is correct) is claimable.

## Current Behavior
The app incorrectly states that `4,662.598321 HNS` is claimable.

This incorrect value can be seen at the following locations:

- https://github.com/kyokan/bob-wallet/blob/1dfd08a59e20fe69426837f789ba7bb869e4bca7/app/pages/GetCoins/index.js#L126
- https://github.com/kyokan/bob-wallet/blob/1dfd08a59e20fe69426837f789ba7bb869e4bca7/app/pages/GetCoins/index.js#L138-L140
- https://github.com/kyokan/bob-wallet/blob/1dfd08a59e20fe69426837f789ba7bb869e4bca7/app/pages/GetCoins/index.js#L181-L194

## Steps to Reproduce
1. Open app
2. Go to the 'Add Funds' page
3. Note that the incorrect value is shown
4. Compare this against the correct value listed at https://github.com/handshake-org/hs-airdrop#how-it-works

## Context (Environment)
N/A

## Logs
N/A

## Screencap
image

image

### LINKED ISSUE #445: Doesn't build on Linux (Ubuntu 20.04)
I followed the [instruction](https://github.com/kyokan/bob-wallet/tree/v0.9.0#linux) on [v0.9.0](https://github.com/kyokan/bob-wallet/tree/v0.9.0) branch and when I do `npm install` i got this error

```
npm ERR! Package libusb-1.0 was not found in the pkg-config search path.
npm ERR! Perhaps you should add the directory containing `libusb-1.0.pc'
npm ERR! to the PKG_CONFIG_PATH environment variable
npm ERR! No package 'libusb-1.0' found
```

it seems `libusb-dev` is version 2 instead of 1

```
$ sudo apt list libusb-dev 
Listing... Done
libusb-dev/focal,now 2:0.1.12-32 amd64 [installed]
```

so i installed `libusb-1.0-0-dev` and run `npm install` again. This time i got an another [error](https://termbin.com/tqrb)
I was using Node 17.2 but downgrading to Node 14 (v14.18.3) with `nvm install 14` (I use [`nvm`](https://github.com/nvm-sh/nvm)), it builds with no issues

### LINKED ISSUE #457: Watching list blank until new name added
2 people reported this:

And everytime I close it, all the names from "watching" tab disappears.

_Originally posted by @DIPMR in https://github.com/kyokan/bob-wallet/issues/429#issuecomment-969031098_

And `namer#2880` on discord:

>My observation after updating to version 0.9: domain watchlist doesn't load - have to add any domain there and then all the previous list is loaded. Always have to do it after restarting BobWallet.

rithvikvibhu:
A bunch of small fixes that don't really need individual PRs. Each commit is independent and only a few lines long.
There are a few more I'd like to add, opening this draft PR so the issues are linked on GitHub and it shows the icon in issues list.

**This must be merged after #456 as that's where the base is branched off from.**

### auction: fix NaN HNS lockup value
(closes #396)
Max lockup from all bid values were not calculated properly and panel had **NaN HNS** lockup (pic in issue).

### auction: replace p with div tag
React throws a large error when there's a div inside a p, so changed the p to a div.

### auction: show open as pending only until it is
(closes #394)
All 6 hours of open period used to say "please wait while OPEN is mined". Now pending OPEN, and waiting for bidding start have different messages / button text.

### domain-manager: hide names in reveal
(closes #378)
Names were listed in Domain Manager even if the reveal period hadn't ended and the wallet was _currently_ the owner with the highest revealed bid. Now, those are hidden.

### txs: update tx history on new blocks
(closes #205)
Transaction history automatically updates on new blocks. The "key" for react's loop over transactions needed to include the pending status, otherwise the transaction component wouldn't update and stays as pending.
There's also a debounce (of 5 secs) when refreshing transactions because it's a bit heavy and rescans / syncing a lot of blocks would parse all transactions each time.

### watchlist: get list on node start
(closes #457)
regression: Watchlist wasn't loaded on start. Now it does on node start (as watchlists are network-specific).

### i18n: extract more hard-coded strings
(closes kyokan#454)
Added more strings to locales json. More info in issue ^.
Note: These strings have been updated in english and will need to be translated again when #446 is merged: `openSuccessText`, `getCoinAirdrop1`

### text: fix airdrop amount, libusb ver
(closes #425, closes #445)
- Updated airdrop amount
- Updated readme to specify libusb version required to build on linux

### onboarding: disallow primary as wallet name
(closes #410)
`primary` is excluded from the wallet list that's used to check if a wallet with the same name exists when creating a new one. So handled it specifically.
This is related to (but does not close) #438 which seems like a better solution. Removing `primary` from all files will need a bigger commit in a separate PR.

### ui: add settings icon on login page
(related to #182)
If an RPC connection is broken without any wallets, there's no way to get back to a working state:
>If Bob switches to use an external hsd node instead of starting its own and it is not able to reach it, it says "no connection".
>Unfortunately, the option to start the internal node is in the Settings page, but it can't be reached until a wallet exists. But it's not possible to create a wallet because there's no connection to a node.

To break this deadlock, allow opening Settings directly from the login page. Only Network & Connection section's direct actions (start/stop node, spv, dns) and custom RPC modal are allowed. Trying anything else goes back to login.

### ui: do not set state if unmounted
In `Account`: `setState` was called after an await, that errored if the component was unmounted before the promise resolved. Now, it verifies that the component is still mounted before updating state.

### throttle updateHNSPrice
The CoinGecko API was being called way more than needed (state updates in Portfolio, etc.)
Now it is throttled to 2 mins.

### ui: no overflow for records
To fix https://github.com/kyokan/bob-wallet/issues/119, overflow was added to collapsibles, but it also affected Records in Domain Manager, where content should not overflow:
image
This is fixed.


rithvikvibhu:
I think this covers all small-change issues.

For anyone who wants to review:
- Checkout this PR, and run, or download pre-built bin: [linux appimage](https://github.com/rithvikvibhu/bob-wallet/suites/5179531917/artifacts/157306144)
- Go through the description text above and see if anything's broken

rithvikvibhu:
Rebased and tested everything again on mainnet.

PR #462, opened 2022-02-06
Allow unencrypted wallets

### LINKED ISSUE #245: Handle unencrypted wallet
When `hsd` is first loaded, a unecnrypted `primary` is automatically created. For old Bob users, we have been simply hiding it from UI, and use the wid `allison` instead.

With new support for multiple account, two issues arise:
1. When opening Bob for the first time, user skipped onboarding and see a dropdown with a `primary` wallet
2. The `primary` wallet can be unlocked by any or no password

This is currently fixed by hiding the `primary` wallet, same treatment as previous versions of Bob.

#### For # 1, it's an issue that will affect **first-time users**

We can fix it by:
- if the wallet ONLY contains an unencrypted `primary` wallet, show onboarding
- default to using `primary` as wid for first-time user
- before creating/importing a wallet, remove the `primary` wallet first

#### For # 2, it's an issue that has to do with **existing users**

We can fix it by:
- if a wallet is encrypted, remove Login input and change CTA to "Open Wallet"
- add persistent "This wallet is not password protected. Set a password here" banner to main wallet view
- add "Change Password" to setting

### LINKED ISSUE #438: Allow "primary" wallet recovery

rithvikvibhu:
Closes #438, closes #245.

When creating a new wallet, if a wallet with the same name already exists and is not encrypted (like the default `primary` wallet), then the flow continues and encrypts it.

Also, since the primary wallet cannot be deleted, it's handled in Settings.

**Note:** Untested with Ledger (watchOnly wallets), but it should work. Probably. Mostly.

#### i18n

2 strings added and need translations:
```
- obSetNameUnencryptedExistError
- removeWalletNoPrimary
```

image


rithvikvibhu:
Had to change a bit of `listWallets` code, and I think it should still work, but can't test with Ledger wallets. Would be great if someone can confirm.

befranz:
I tested following steps with this PR (on regtest):
✅ Create new wallet with Ledger Nano X
✅ Logout/Login this wallet

Issue #324, opened 2021-04-06
[Feature][Bounty] HIP-0002


Falci:
Some time ago, I did [a PoC where Bob Wallet could send HNS to a name](https://youtu.be/famEYy8fV4A).

Later, we formalised the process in the [HIP-0002: Well Known directory for wallets address](https://hsd-dev.org/HIPs/proposals/0002/).

Even [a small js lib](https://www.npmjs.com/package/well-known-wallets-hns) was built to help making secure requests.

But this was never finished. 😕

The video shows the "Send" page, but it should also be possible to apply this feature in the Transfer page.

Important to note: we need to come with a solution for conflicts: what happens if someone has a domain name: `hs1q...`? How can we know if the user is trying to transfer to an address or to a domain?

I'd like to offer **200 USD** (in HNS) as bounty to see this feature done.

**Requirements:**
* Implement the HIP-0002;
* It should only use https requests;
* This feature should be applied to the Send and Transfer page;
* It should have a way to solve the name/address conflict;

skyinclude:
I'm doing a video about this right now on the SkyInclude channel!  Let's get more people behind this! 

It is all about community! 

thanks for all you do for the protocol, Falci!

pinheadmz:
> Important to note: we need to come with a solution for conflicts: what happens if someone has a domain name: `hs1q...`? How can we know if the user is trying to transfer to an address or to a domain?

@Falci this might be something we could add as rule to HIP2: if the domain name string is a valid HNS address, the protocol fails and no HIP-2 address is returned. For all coins BTC and ETH etc this rule should apply.

rithvikvibhu:
Can we also add a requirement (or at least recommendation) that the logic is placed in a separate js library and then used in Bob?
I'm sure a lot of new services will also want to use it and will make that easy - just like how `.eth` links are supported in many wallets, exchanges, apps.

Falci:
> @Falci this might be something we could add as rule to HIP2: if the domain name string is a valid HNS address, the protocol fails and no HIP-2 address is returned. For all coins BTC and ETH etc this rule should apply.

IMO, it could be solved with UX: the field expects a valid HNS address, and we could have a button next to it which opens a modal dialog to input the domain name. This last could be even something similar to a valid address, as long as it follow the HIP2.

> Can we also add a requirement (or at least recommendation) that the logic is placed in a separate js library and then used in Bob?
[...]

PR #465, opened 2022-02-08
HIP-0002

lukeburns:
Following up on https://github.com/kyokan/bob-wallet/issues/324

I've published two libraries which this PR depends on (based heavily on @Falci's code):
1. https://github.com/lukeburns/https-dane
2. https://github.com/lukeburns/well-known-wallets-dane

To send HNS to a well-known address, you add an `@` before the domain, e.g. `@iamfernando`, and it's resolved using `well-known-wallets-dane`.


lukeburns:
Latest currently here: https://github.com/lukeburns/bob-wallet/tree/update-deps

Waiting for https://github.com/kyokan/bob-wallet/pull/444 to merge

app/components/SendModal/index.js
```diff
@@ -192,13 +252,25 @@ class SendModal extends Component {
...
             <div className="send__label">{t('sendToLabel')}</div>
             <div className="send__input" key="send-input">
+              {hip2Input && <span className="send__prefix">{to ? (
+                <img src={LockSVG} />
+              ) : '@'}</span>}
               <input
                 type="text"
-                placeholder={t('recipientAddress')}
+                className={hip2Input ? (to ? 'send__input-hip2-success' : 'send__input-hip2') : null}
+                placeholder={hip2Input ? t('recipientHip2Address') : t('recipientAddress')}
```
rithvikvibhu, line 261: Do you think it's a good idea to indicate in the placeholder that HIP-2 resolution is enabled and possible?
Like instead of just `Recipient address` as default, it could say `Recipient address (start with @ for alias)` or similar.

app/components/SendModal/index.js
```diff
@@ -54,25 +64,75 @@ class SendModal extends Component {
...
-    this.setState({to: e.target.value, errorMessage: ''});
-    if (e.target.value.length > 2 && !isValidAddress(e.target.value, this.props.network)) {
-      this.setState({errorMessage: this.context.t('invalidAddressPrefix')});
+  openLinkHandler (e) {
+    e.preventDefault()
+    shell.openExternal(e.target.href)
+  }
+
+  updateToAddress = async e => {
+    const dnsEnabled = !this.props.noDns
```
rithvikvibhu, line 90: Didn't really occur to me that DNS Server must be enabled in settings for HIP-2 resolution to work until I read the code. Is there any way we can tell users this in ui?

rithvikvibhu: Also, currently, DNS servers aren't enabled by default. Should we change `noDns` default from `yes` to `no`? i.e. enable dns by default.

app/components/SendModal/index.js
```diff
@@ -192,13 +252,25 @@ class SendModal extends Component {
...
                 onChange={this.updateToAddress}
-                value={to}
+                onKeyDown={this.updateHip2}
+                spellCheck="false"
+                value={hip2Input ? hip2To : to}
               />
             </div>
+            { hip2Input && (
+              <React.Fragment>
+                <Alert type="error" message={hip2Error} />
```
rithvikvibhu, line 270: This error stays as the domain is typed out. It's kinda hard to make out if the domain I've just typed really doesn't work, or if it isn't resolved yet. Can we clear the error soon as input changes (`onChange`)?
Even better, some kind of a tiny loading spinner if possible.

app/components/SendModal/index.js
```diff
@@ -54,25 +64,75 @@ class SendModal extends Component {
...
+
+    const address = this.state.to
+    if (address.length > 2 && !isValidAddress(address, this.props.network)) {
+      this.setState({ errorMessage: this.context.t('invalidAddressPrefix') });
     }
   };
 
+  updateHip2 = e => {
+    if (this.state.hip2Input) {
+      if (e.key === 'Escape' || (e.key === 'Backspace' && this.state.hip2To.length === 0)) {
```
rithvikvibhu, line 130: Hey nice, I like that even `esc` is captured and works :D

rithvikvibhu (changes requested):

rithvikvibhu:
Hey, just tried it out and everything works great functionally, nice work @lukeburns! Just a few review comments, mostly small suggestions.

One thing I wanted to see what your thoughts are on disabling HIP-2 resolution until chain sync is complete. Because of 2 reasons:
- Old / Incorrect NS records or recirds from an old owner could be queried (not sure if the DNS server just servfail's or returns old records)
- When in SPV and syncing, no dns records are served at all, even for ICANN domains like `falci.me`

Disabling until chain sync (in both full/spv) addresses both points afaik, or you might have a better solution for them.

app/components/SendModal/index.js
```diff
@@ -192,13 +252,25 @@ class SendModal extends Component {
           <div className="send__to">
             <div className="send__label">{t('sendToLabel')}</div>
             <div className="send__input" key="send-input">
+              {hip2Input && <span className="send__prefix">{to ? (
+                <img src={LockSVG} />
```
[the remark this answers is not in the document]
Falci: I vote for displaying the 🔒

package.json
```diff
@@ -146,6 +146,7 @@
     "copy-to-clipboard": "3.3.1",
     "deep-equal": "2.0.1",
     "electron-debug": "3.2.0",
+    "hip2-dane": "^0.2.0",
```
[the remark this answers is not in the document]
Falci: What you mention is pretty important. But would it matter if `hip2-dane` have other dependencies that uses `^` ?

https://github.com/lukeburns/hip2-dane/blob/17fa8e23405bda45a8c771ca94a12b22c12a5160/package.json#L15

app/components/SendModal/index.js
```diff
@@ -13,6 +13,15 @@ import { clientStub as aClientStub } from '../../background/analytics/client';
...
 import {I18nContext} from "../../utils/i18n";
+import LockSVG from '../../assets/images/lock.svg';
+import hip2 from "../../utils/hip2Client";
+
+const networkPorts = {
+  'main': 5350,
+  'testnet': 15350,
+  'regtest': 25350,
+  'simnet': 35350
+}
```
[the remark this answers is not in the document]
rithvikvibhu: Oh true, yes those deps too, nice catch.

package.json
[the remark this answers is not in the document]
rithvikvibhu: Oh true, yes those deps too, nice catch.

app/components/SendModal/index.js
```diff
@@ -54,25 +60,110 @@ class SendModal extends Component {
...
-    this.setState({to: e.target.value, errorMessage: ''});
-    if (e.target.value.length > 2 && !isValidAddress(e.target.value, this.props.network)) {
-      this.setState({errorMessage: this.context.t('invalidAddressPrefix')});
+  openLinkHandler (e) {
+    e.preventDefault()
+    shell.openExternal(e.target.href)
+  }
+
+  updateToAddress = async e => {
+    const hip2Enabled = !this.props.noDns && this.props.isSynchronized // && !this.state.simulateSynchronizing
```
rithvikvibhu, line 88: Guess `&& !this.state.simulateSynchronizing` is left out from testing? :wink:

app/components/SendModal/index.js
```diff
@@ -54,25 +60,110 @@ class SendModal extends Component {
...
+  }
+
+  handleEscape = e => {
+    if (this.state.hip2Input && e.key === 'Escape') {
+      this.resetInput()
+    }
+
+    // if (e.metaKey && e.key === 's') {
+    //   this.setState({ simulateSynchronizing: !this.state.simulateSynchronizing }) // toggle sync simulation
+    // }
```
rithvikvibhu, line 156: and these

rithvikvibhu:
Not blocking for this PR, these strings need to be translated:
```
hip2DisabledWarning
hip2InvalidTLSA
noHip2AddressFound
recipientAddressHip2Enabled
recipientAddressHip2Syncing
recipientHip2Address
```

lukeburns:
Default DNS setting is here for my own future reference https://github.com/kyokan/bob-wallet/blob/b9d1601cadee24e257dfc13e81ce13c66a3edba0/app/ducks/nodeReducer.js#L30

app/components/SendModal/index.js
```diff
@@ -178,27 +269,68 @@ class SendModal extends Component {
...
+              <span className="send__prefix">{to ? (
+                <img src={LockSVG} />
+              ) : ( hip2Loading ? 
+                <img className="send__hip2-loading" src={RingsSVG} /> 
+                : '@')}
+              </span>}
               <input
                 type="text"
-                placeholder={t('recipientAddress')}
+                className={hip2Input ? (to ? 'send__input-hip2-success' : 'send__input-hip2') : null}
```
rithvikvibhu, line 310: This className doesn't have any styles, can be removed?

rithvikvibhu:
I wasn't able to get the gray bordered syncing alert to show (but placeholder changed as expected). When does it show up exactly?

Other than this and and the DNS server setting, it LGTM!

lukeburns:
The gray bordered thing shows up when the node starts syncing while you’re doing an alias lookup. I was testing this case with the commented out simulate synchronization code.

rithvikvibhu (approved):

lukeburns:
Changes:
1. Root and recursive nameservers enabled by default and listening on non-standard ports, 9891 and 9892 respectively.
2. Added a section in Settings to edit hip2 port, that can be set to resolve aliases using e.g. fingertip (port 9592) or the system resolver (port 53).

rithvikvibhu:
I'm seeing this error in the console when Bob runs now:
```
[128309:0221/194842.958498:ERROR:CONSOLE(1)] "Uncaught TypeError: Cannot read properties of undefined (reading 'instance')", source:  (1)
```
Not sure where it's from (no stacktrace). Any idea where it's coming from?

lukeburns:
Hmmm, I'm not sure where this is coming from, but it must be coming from a dependency, because I'm can't find any property `.instance` in the codebase.

lukeburns:
For future reference
https://github.com/electron/electron/issues/32768 https://github.com/electron/electron/issues/32702

Nathanwoodburn (approved):
This is super useful as it is heaps easier to remember a domain (eg. nathanwoodburn) than an address (eg. hs1qvxhtxl0eq363wp68aqgslr93ufcly5r3kc3e34)

I can't wait for it to be added as default.

rithvikvibhu:
@lukeburns last one I promise :|

Tried out a couple of unusual cases.

### Website returns a non-Handshake address
Test case: `invalid-addr.bb-8`
image

Bob should make sure that it is a valid HNS address. Easiest would be using `Address` from hsd. Also, it might be a good idea to not show the response content if it is not an address - just something like "Invalid Address" or similar.

### Website returns binary / different content type
Test case: `invalid-type.bb-8`
image

I used svg here, but applies to binary too. Better to not show content, just a constant error.

### Website returns a very large response
Test case: `large-size.bb-8`
image

This actually downloads the whole mp4. I'd suggest replacing https://github.com/lukeburns/hip2-dane/blob/4bbd57ba1922edc5b0af094203b8107165bb2073/hip2-dane.js#L12 with:
```js
res.on('data', chunk => {
  if (data.length > 1000) { // 1 KB
    req.destroy();
    return reject('response too large')
  }
  data += chunk
})
```

app/ducks/hip2Reducer.js
```diff
@@ -0,0 +1,26 @@
...
+}
+
+
+function getInitialState() {
+  return {
+    port: 9892
+  }
+}
+
+export default function hip2Reducer(state = getInitialState(), action) {
```
rithvikvibhu, line 19: Packaged builds no longer start (try it out with `npm run package-linux` or whichever os you're running).

`ducks/nodeReducer.js` needs to be added to `scripts/package.sh` as well as `scripts/package.js`.

rithvikvibhu (changes requested):

lukeburns:
Alright, have length + prefix address validation working. If it is a bad encoding, Bob catches it when you try to confirm payment.

Screen Shot 2022-03-07 at 12 30 45 PM

lukeburns:
https://github.com/lukeburns/hip2-dane/blob/main/hip2-dane.js#L20-L33

`hip2-dane` now only grabs the first <= 90 characters up to the first new line.

lukeburns:
- updated hip2-dane to invalidate aliases that are valid addresses
- now validating addresses based on current network type
- fixed a bug preventing input while syncing

rithvikvibhu (approved):

PR #469, opened 2022-02-11
SPV, RPC, Rescan fixes

### LINKED ISSUE #310: Auction rescan broken when using custom RPC, (0.6.2)
## Expected Behavior
Auction rescan should work when using custom RPC 

## Current Behavior
When using custom RPC even after rescanning an ongoing auction no bids are displayed in Bob.

## Steps to Reproduce
1. Connect to a custom RPC node
2. Rescan any ongoing auction with bids 
3. See that no bids are displayed.

## Context (Environment)
Trying to use an external node with Bob instead of the internal one.

## Logs
N/A

## Screencap
image

image

## Additional Info
In a discussion on telegram, @pinheadmz said `the "rescan auction" function calls wdb.rescan() which may have new functionality in bob, i dunno I think depending on your version the local rescan method has been stubbed in` and @chikeichan said this was likely the problem.

### LINKED ISSUE #421: Status stuck on rescan in SPV mode
When in spv mode, clicking on **Rescan Auction** starts rescan and completes, but the status gets stuck on **Rescanning... 100%**. The auction page doesn't let you place bids until you leave and return to the page.

The stuck status doesn't change even on new blocks / height. Left sidebar "Current Height" always says `XX/XX` with XX being the current height (which does change on new blocks).

Bob 0.9.0-rc1, Manjaro / Linux.

### LINKED ISSUE #466: "generate to address" buttons in regtest won't work with SPV node
We should hide them in SPV mode. Also, I think in SPV mode there should be some global indicator on the UI (like by the sync %)

### LINKED ISSUE #467: Needs app restart on toggling SPV mode
Screen Shot 2022-02-08 at 1 18 08 PM

Switching to SPV from this screen did lock up the app though. I had to restart it twice to get it to start syncing SPV. One thing that may be an issue swithcing from full node to SPV is the hosts.json file. I find that deleting this file gives the spv node a much better chance of finding supporting peers.

_Originally posted by @pinheadmz in https://github.com/kyokan/bob-wallet/issues/463#issuecomment-1032924871_

### LINKED ISSUE #515: Add the rig options on the loading screen (Working on it)
Add the rig options on the loading screen (Working on it)

During boot app, add the "rig" option button to be able to go to settings page,
When you use Remote HSDS Node in settings, then you shut down the VPS, the next time you open Bob, it will stuck and you will not be able to change to other options such us: use SPV node or Start node, so you will never be able to use Bob.

Miguel Gargallo is the CEO of Pencil Domains, pencil domains is the 1st Europe Andorra Canarias Spain Registrar  Miguel Gargallo is telling to the world new fix for Bob Wallet development

rithvikvibhu:
SPV, RPC and rescan fixes. Redid commits to make review easier.

### node: set spv boolean correctly
- closes #467
- Now there's no need to restart Bob multiple times when toggling SPV.

### ui: hide generatenewblocks if spv
- closes #466
- SPV can't generate blocks, hides the +1, +10, +50 buttons

### node: remove getAuctionInfo
- Get Auction Info is a wallet call, but was part of node service, removed it

### node: common api url constant
- A single constant for URL for handshake api (remote node used in SPV mode) - easier to replace during dev

### exchange: disable in spv mode
- Related to #480
- Disables Exchange page in SPV with a message. Keeping the issue open until shakedex is updated to use Bob's node service including remote api calls

### rpc: allow http for private IP addresses
- From Discord, a user wanted to use Bob over RPC to a node on a local network.
- Instead of allowlisting just localhost, it now checks and allows HTTP for private IP ranges too (hostnames still require HTTPS, even those that resolve to private IP addresses).
- Uses `binet` (added to package.json, but hsd itself requires it so doesn't add anything)

### rpc: splash screen switch to internal node on error
- closes #515
- If Bob is set to use a remote RPC node, and the node is not reachable, Bob errors on splash screen
- If the node cannot be started (say the IP address is released, or domain is not under user's control), then the only way to get Bob working again was to clear all Bob settings by deleting the `db` directory.
- Now, if it's an RPC connection failure, the splash screen has a button to switch to internal node

### node: throttle refreshNodeInfo
- `refreshNodeInfo` is a slightly expensive call (getInfo and getFees) and was being called for every new block, which is very fast in SPV mode
- It is now throttled.

### wallet: replace rescan hooks with block connect events
- closes #421, closes #310
- depends on https://github.com/handshake-org/hsd/pull/707, https://github.com/handshake-org/hsd/pull/752
- The biggest change of this PR, replaces the rescan unhook and unhooking mitm with wdb `block connect` events.
- Simplifies rescan logic that works well with Full, RPC and SPV (thanks @pinheadmz for https://github.com/kyokan/bob-wallet/pull/469#issuecomment-1079254461)
- SPV (active) auction rescans requires https://github.com/handshake-org/hsd/pull/752. Why it affects us (read this PR desc for `(1)` and `(2)`:
>- Before requesting blocks, a node asks for peer's mempool (but only in regtest!)
>- We know that sendGetBlocks doesn't really request anything, so nothing happens after that (2)...
>- BUT, because of (1), the node expects a mempool inv response - which it won't get because mempool is empty - and after 30 secs, kills the connection.
>- With only set to a single peer, it connects to the same peer again and now makes getBlocks for "the first time" and gets the new blocks.
>To a Bob user, it'll look like a rescan took >35 seconds.

Now, rescans work everywhere and show proper status (next commit below):
|  | Settings -> Rescan | Auction Rescan |
|---|---|---|
| Full node | ✅ | ✅ |
| SPV node | ✅ | ✅ |
| RPC - Full node | ✅ | ✅ |

### wallet: correct resync progress in spv
- resyncing progress in sidebar and topbar now use resync height added in previous commit

### pkg: update to hsd v4.0.1
- Updates `hsd` to v4.0.1 which has https://github.com/handshake-org/hsd/pull/752 merged

---

Overall, SPV seems more stable now, and syncs from scratch in 25 minutes.

TODO:
- [x] Fix tx history not refreshing on rescan complete
- [x] Resolve all "TODO" comments
- [x] Import names in spv
- [x] Update hsd version after https://github.com/handshake-org/hsd/pull/707 is merged
- [x] Remove console.logs


pinheadmz:
👍 looks good so far.

pinheadmz:
@rithvikvibhu re: https://github.com/kyokan/bob-wallet/pull/469/commits/7af427ae6b121a28baaa788e010b6781bd7dc0c6 

I think we can make this a lot simpler by adding a feature to hsd: https://github.com/handshake-org/hsd/issues/621

If wallet and node both emit "new chain height" events, we will always know their sync status and we can drop the `unhook` hack and forget about "wallet rescan" events entirely. This should work for SPV fine as well.

https://github.com/kyokan/bob-wallet/blob/de90b37e9955350c58f1b92af7da38ece1ec0571/app/background/wallet/service.js#L203

rithvikvibhu:
@pinheadmz using hsd fork (https://github.com/handshake-org/hsd/pull/707) and yeah it made things simpler!

SPV importname is still a problem though.

Issue #180, opened 2020-07-26
[Feature] App Version Popup, Cleanup


sdtsui:
App version does not appear to be visible on all builds - see chat on 07/25/2020 for an example of 4.0/Windows.
We should just have an about popup we can show from the menubar.

Cleanup refers to unnecessary links in windows top-menubar that point to electron documentation.
Potentially better to link to Handshake-related docs/HSD if nothing else.

Issue #143, opened 2020-06-05
Windows: help menu is electron stuff


pinheadmz:
![image](https://user-images.githubusercontent.com/2084648/83910525-ba127380-a738-11ea-8509-3b6cfea7445a.png)

Also can't seem to find the version we added in #124 but I think part of that PR was understanding that it would only apply to OSX?

rithvikvibhu:
On non-macos, it looks fine I think:
- File -> Close
- View -> Toggle Full Screen
- Help -> About Bob, Report an Issue

There's code specific to macos that's different and with more items (can't test it though): https://github.com/kyokan/bob-wallet/blob/master/app/menu.js

Can we close if it's fine? (didn't see any electron links)

rithvikvibhu:
Closed by #481.

PR #473, opened 2022-02-20
Change the report url to github new issue link

eskimo:
Since the previous link doesn't work, I think the github new issue link is a good replacement.

### DIFF
--- a/app/components/Notification/index.js
+++ b/app/components/Notification/index.js
@@ -80,7 +80,7 @@ export default class Notification extends Component {
         <div
           className="notification__issue-wrapper__action"
           onClick={() => {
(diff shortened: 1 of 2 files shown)


rithvikvibhu:
Thanks @eskimo for the PR!

closes #143, closes #180, LGTM.

Falci:
Didn't see your PR before.

We could define the links in a single place and reference them wherever we need.
My first idea was to define them in the `package.json`. 

WDYT?

eskimo:
I know nothing about this project at all, so changing the links how they were originally defined is basically the best I can do :P

rithvikvibhu:
#481 replaces this and uses `package.json` values. Thanks again for this PR @eskimo!

PR #477, opened 2022-02-27
Place multiple bids on a name in same block

### LINKED ISSUE #179: [Feature / Bug] UI should prevent double-submitting bids, but allow concurrent pending bids
Copied from discussion of #164 

>> ANOTHER followup/related issue is that when there's a bid pending, I can't attempt another bid, even though I might really want to (i.e. if a sniper bid appears in the mempool right after I submitted the bid, on the last block of bidding), without deleting all pending tx. It'd be nicer if I could just go ahead and bid as often as I want, perhaps with a little warning text about the pending ones and perhaps a confirmation click to avoid accidental double-tap bids. It would also be nice to be able to delete the pending tx more selectively by auction for the same purpose.

> Definitely agree with this -- could you open up a new issue / feature request? I think an implementation would just need to prevent users from accidentally sending the same bid twice, so we need obvious feedback that their currently pending bid does not need to be re-sent, that "pending" is OK.

rithvikvibhu:
Closes #179.

Demo (https://youtu.be/FNT78BhzhS4):
[YouTube video](http://www.youtube.com/watch?v=FNT78BhzhS4 "Place multiple bids on a name in a single block")

Changes made to make this happen:
- Add `getBlind` to `walletClient`
- While parsing pending transactions,
  - store an array in case of BID, leave other covenants as before
  - get true bid value from blind from output covenant item
- reset input fields for true bid and blind in UI after sending a bid
- change "place bid" button text to "place another bid" to emphasize that a bid has already been placed
- `BidHistory` component to show pending bids differently (gray text, italics)
- Move `reducePendingTransactions` from `namesReducer.js` to `walletActions.js` because it no longer "just" reduces, it also calls async function getBlind. It anyway did a lot more than a reducer should before also.

Also:
- Update auction details / name info / bids on new blocks - no need to refresh screen or navigate away and come back anymore
- Extract strings from modal on placing bids for i18n

Recommend squash and merge because commits aren't really separate.


pinheadmz:
Do you think we should alert the user if they are placing a duplicate bid?

rithvikvibhu:
I initially thought that the big blue "placed bid" modal and the button changing text is enough, but going through the video again, yeah an explicit warning might be better.

rithvikvibhu:
@pinheadmz added that warning when one tries to bid again with the same value:
image

brandondees:
might be a nice tweak to also extend the "another bid" copy change to all the other submit/confirm parts of the UI, not just the first button. it would be easy to open the dialogue, get distracted, then come back and forget that the open prompt is for an extra bid, especially if the first bid's state may have changed in a way the user was unsure about in the meantime.

app/pages/Auction/BidHistory.js
```diff
@@ -33,14 +34,17 @@ export default class BidHistory extends Component {
 
     bids.forEach(bid => {
       order.push(bid.from);
+
+      const {month, day, year} = createAMPMTimeStamp(bid.date);
       map[bid.from] = {
-        date: createAMPMTimeStamp(bid.date),
+        date: bid.date ? `${month}/${day}/${year}` : '(pending)',
```
brandondees, line 40: we'll probably have to revisit date/time formatting in I18n friendly way at some point. not important to focus on here, just observing

app/pages/Auction/index.js
```diff
@@ -106,6 +107,17 @@ export default class Auction extends Component {
...
+  async componentDidUpdate(prevProps, prevState) {
+    if (this.props.chain.height !== prevProps.chain.height) {
+      this.refreshInfo();
+    }
+  }
+
+  refreshInfo = throttle(async () => {
+    await this.props.getNameInfo(this.getDomain());
+    await this.props.fetchPendingTransactions();
+  }, 10*1000, {leading: true}) // 10 seconds
```
brandondees, line 119: is this throttle rate based on some timeout defined elsewhere, or just an arbitrary rate?

does throttle handle the situation where the async function runs longer than 10 seconds? i.e. if it took 60 seconds to complete, would it fire up to 6-ish concurrent attempts, or would it wait for the first one to resolve before allowing the second to begin?

also, if i'm reading correctly, given the componentDidUpdate is async, and this function awaits on each line, is the async inside of throttle actually providing any benefit?

brandondees (approved):
I read through this PR and it looks like a really nice improvement. I added a couple non-blocking comments.

app/pages/Auction/index.js
rithvikvibhu: Good points. 10 seconds just seemed enough, since `fetchPendingTransactions()` is a local call and `getNameInfo` is small json and fixed in size (if local or even if remote over network).
I still don't like the idea of random numbers like this, but without a throttle, it will refresh every time the height changes (which can happen more than once a second when syncing). Open to other cleaner ideas.

>given the componentDidUpdate is async, and this function awaits on each line

You're right, we don't really care when it finishes and 118 doesn't need 117 to finish first, so removed await (and async) so they can run in parallel.

Issue #436, opened 2021-12-24
GPG Signature to Verify Download


BTCAlchemist:
Hello, I would like to verify the download with a verified GPG signature since the checksum hash could be manipulated. I may have missed this information. Is the latest release (0.8.0 Stable) signed with GPG by any of the lead devs (and what is their GPG fingerprint or signature)? Thank you in advance.

rithvikvibhu:
Will start signing from the next release with (at least) 1 key :)

For now, v0.9.0:

```
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA256

49d5273d1fccf29f4bbc63f0a3acd059366d4cebe75f5d8573381ebbfff831e885c9f07fa56adadca416fffd703f84291b7d8c2b78b9b86945ee8b3478799c3f  Bob-0.9.0.AppImage
e57fb88c0b0d930e2fbef29fdf27c1765ce2d6ca83b2fcb9af4963d853056cf34ac813292754cc4482259005c23295990048a32a4f440049f2bd9a73ea20d286  Bob-0.9.0.dmg
028e41169cf7987f57aa0207c2487bcb3bea7129a4264bea76535945a1ee9ca53f3f906f61bd0420a4b3b8afdb4a4fdf8e0bd6fbd426103073ad84966ff6f994  Bob.0.9.0.msi
-----BEGIN PGP SIGNATURE-----

iQEzBAEBCAAdFiEES9ppUueR+VbP+LbSkN+sghCpF9kFAmH/hh4ACgkQkN+sghCp
F9l7tQgAqCVS5m9kTdlac1+Dvksd5BEIwl3510ip6bxunx0eN8gMcrAbr/2difYF
pLqr2FoAFwotx1qX/4UBnVb99tRN64nLgsbPJ3nGT5syAQZtKhP2ajuV4KREu4y1
62hc0ZryrM095niGN2o9LdpVS/cTWK17gIvD+fmV77sNwE6ns2i75QqIg8vOgrAQ
v8mlOYpYXwFahdcp+4c7oArC/Kj+vU3qO+7xI3hNG1e+qX7bdAa8AxbVvK4IVaM3
HD70ZSntnS6ydkl9joQHXPgN+AKdZlWMF1Ac1t5Jw7pt42ybLWroGJ4Jt0Vaiiva
VBiaR9WFwqm38TT4XYxko7mb6OQvUA==
=6oNn
-----END PGP SIGNATURE-----
```

[...]

PR #478, opened 2022-02-28
Update notification

rithvikvibhu:
Checks for updates on Bob Wallet start, and if a new stable release is found, shows a notification in sidebar and clicking on it opens the GitHub release page for that version:

image

Uses `semver` to compare, so RCs and others are handled properly. Just need to make sure that release tags are in format `v${version}` (like `v0.9.0`), which we've already been doing.

### DIFF
--- a/app/background/setting/service.js
+++ b/app/background/setting/service.js
@@ -37,6 +37,21 @@ export async function setCustomLocale(json) {
   return await put(CUSTOM_LOCALE, JSON.stringify(json));
 }
 
+export async function getLatestRelease() {
+  try {
+    const releases = await (
+      await fetch(
+        'https://api.github.com/repos/kyokan/bob-wallet/releases'
+      )
+    ).json();
+    const latest = releases.filter(r => !r.draft && !r.prerelease)[0];
+    return latest;
+  } catch (error) {
+    console.error(error);
+    return null;
+  }
+}
+
 const sName = 'Setting';
 const methods = {
   getExplorer,
@@ -45,6 +60,7 @@ const methods = {
   setLocale,
   getCustomLocale,
   setCustomLocale,
+  getLatestRelease,
 };
 
 export async function start(server) {
--- a/app/ducks/app.js
+++ b/app/ducks/app.js
@@ -28,7 +28,7 @@ export const checkForUpdates = () => async (dispatch) => {
       type: SET_UPDATE_AVAILABLE,
       payload: {
         version: latestRelease.tag_name,
-        url: latestRelease.html_url,
+        url: `https://github.com/kyokan/bob-wallet/releases/tag/${latestRelease.tag_name}`,
       },
     });
   }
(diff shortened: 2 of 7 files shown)


pinheadmz:
Concept ACK! Would it be crazy to help users verify the PGP sig on new dowloads as well? I dunno if that makes sense, theoretically a compromised verion could convince a user to download additional compromised versions.... but maybe there is something we can do? Im curious if other crypto-sensitive software handles releases in any special way like that.

rithvikvibhu:
>Would it be crazy to help users verify the PGP sig on new dowloads as well?

I think we can add more info in release docs and on the website on how to verify checksum and gpg easily (#436, #214).

Making the software itself verify an upgrade will complicate things imo - this would mean moving the download/install/update process inside the wallet which is messy when cross-platform.

app/background/setting/service.js
```diff
@@ -37,6 +37,21 @@ export async function setCustomLocale(json) {
   return await put(CUSTOM_LOCALE, JSON.stringify(json));
 }
 
+export async function getLatestRelease() {
+  try {
+    const releases = await (
+      await fetch(
+        'https://api.github.com/repos/kyokan/bob-wallet/releases'
```
pinheadmz, line 44: Mine as well check some kind of signature, even if its a brand new bob-release-update published by kyokan. Especially because you use `url: latestRelease.html_url,` down below. Giving users a link to click is a target for attacks, treat it like a nuclaar bomb.

https://bitcoinist.com/electrum-wallet-phishing-bitcoin/

rithvikvibhu: Wow didn't think of that. Unlikely, but def possible.

How about always linking to `https://github.com/kyokan/bob-wallet/releases/tag/{TAG}` instead of `latestRelease.html_url`?

Not sure what signature can be checked here. Do you mean like with PGP? It's just a link so no info to verify.

rithvikvibhu:
@Falci heh. Releases aren't really frequent (2 per year?). Having a setInterval seems like overkill imo. Wdyt?

app/background/setting/service.js
rithvikvibhu: Updated to prefix url and only use tag. imo the only way to break this would be to MITM a HTTPS req to github.com, which if someone can do, there are bigger problems already.

>maybe an HNS address so we can use rpc verify message

Requires the node to be synced :|

>I dunno if it's possible to check that the release signature on github is correct before notifying the user.

Afaik, no. We could download another file and check that, but again relies on an external source, mostly over https, with the same attack vectors as https://github.com. Additionally, verifying a file would mean downloading the actual binary from within Bob which feels more dangerous.

pinheadmz: > Requires the node to be synced :|

no, just the address, not the NAME.

app/ducks/app.js
```diff
@@ -28,7 +28,7 @@ export const checkForUpdates = () => async (dispatch) => {
       type: SET_UPDATE_AVAILABLE,
       payload: {
         version: latestRelease.tag_name,
-        url: latestRelease.html_url,
+        url: `https://github.com/kyokan/bob-wallet/releases/tag/${latestRelease.tag_name}`,
```
pinheadmz, line 31: Yeah, good change.

PR #486, opened 2022-03-21
node: use different wallet api key

### LINKED ISSUE #484: Can not set custom API key
to reproduce:

- open bob
- go to settings > wallet
- set a custom API by typing in the text area
- restart bob (the modal says to restart for new api key to take effect)
- go back to settings and look at api key, its been randomly regenerated

rithvikvibhu:
Closes #484.

Instead of using node's api key for wallet also, first check if wallet has a custom api key in db already.

New behavior:
- On fresh start, node and wallet get a newly generated random api key (same for node and wallet)
- (wallet as plugin) node or wallet api key can be individally changed in settings without affecting the other

RPC is not modified, it just uses the custom wallet api key if stored in db or generates (and stores) one.

### DIFF
--- a/app/background/node/service.js
+++ b/app/background/node/service.js
@@ -12,11 +12,6 @@ import { ConnectionTypes, getConnection, getCustomRPC } from '../connections/ser
 import FullNode from 'hsd/lib/node/fullnode';
 import SPVNode from 'hsd/lib/node/spvnode';
 import plugin from 'hsd/lib/wallet/plugin';
-import Address from 'hsd/lib/primitives/address';
-const blake2b = require('bcrypto/lib/blake2b');
-const secp256k1 = require('bcrypto/lib/secp256k1');
-const {safeEqual} = require('bcrypto/lib/safe');
-import pkg from 'hsd/lib/pkg';
 import { prefixHash } from '../../db/names';
 import { get, put } from '../db/service';
 import {dispatchToMainWindow} from "../../mainWindow";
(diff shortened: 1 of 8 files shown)


pinheadmz:
I'm trying out this branch locally now and although I can set a custom API key for the wallet that persists after restart, the node API key remains the random one generated by bob on first boot. Luckily, that original key persists after restart but there is no way to view it again from settings.

Why can't there just be ONE API key for both node and wallet, and if the user sets a custom one, it applies to both and doesn't change until the user commands that?

Also, I'm not totally sure the UX around setting a new key is right. I enter the new key and click "update" which greys out but there is no further feedback. I think the message "you must restart to take effect" should pop up in red when you click "update" in that modal, maybe even offer a "shut down now" button ?

rithvikvibhu:
>Luckily, that original key persists after restart but there is no way to view it again from settings.

wdym? Node and wallet API keys can both be seen in settings any time.

>Why can't there just be ONE API key for both node and wallet

Would this affect RPC? Where you'd possibly have different (remote) node and (local) wallet keys.

>I think the message "you must restart to take effect" should pop up in red when you click "update" in that modal, maybe even offer a "shut down now" button?

Yeah good idea. Think we can even go ahead and restart the node on key update, like how the DNS on and off works.

pinheadmz (approved):

rithvikvibhu:
Now the node restarts as soon as "Update" is pressed and the modal closes. (also rebased for merge)

PR #431, opened 2021-11-26
Broadcast Bob's name and version via hsd

Falci:
This change allows Bob wallets to announce themselves to their hsd peers.

Depends on 
~~handshake-org/hsd/pull/664~~
handshake-org/hsd/pull/710

PR #521, opened 2022-06-16
Upgrade hsd to v4.0.0 with chainDB migration and tree compaction

pinheadmz:
```
[I:2022-06-16T20:41:31Z] (chain) Compacting Urkel Tree...
[I:2022-06-16T20:45:29Z] (chain) Synchronizing Tree with block history...
[I:2022-06-16T20:46:35Z] (chain) Synchronized Tree Root: 8b58e5f725336accbb1775ef0d1a8dea542a8d2c9d31ecb8cadabc4db73fee67.
```
Took my macbook pro about 5 minutes

Screen Shot 2022-06-16 at 4 50 02 PM

### DIFF
--- a/app/background/node/service.js
+++ b/app/background/node/service.js
@@ -23,6 +23,7 @@ import {
   SET_SPV_MODE,
   START_NODE_STATUS_CHANGE,
   END_NODE_STATUS_CHANGE,
+  COMPACTING_TREE,
 } from "../../ducks/nodeReducer";
 import pkg from '../../../package.json';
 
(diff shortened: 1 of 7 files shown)


pinheadmz:
Note that the actual migration only takes like 1 second, but the tree compaction takes several minutes. The way we have it configured by default, every 10,000 blocks (about 2-3 months) when the user restarts, it will compact again.

If we think this time period should be extended, we can add another setting to the full node
`compactTreeInitInterval: <blocks>`

pinheadmz:
also updates hsd-ledger to 2.0.0 from npm instead of github, and no longer requires `--legacy-peer-deps`

rithvikvibhu (approved):
Everything was smooth, still unbelievable how much space this frees! Took ~6 minutes for `tree` to be deleted and replaced by `tree~` and another 2 minutes for Bob to start. Could see the login screen in 8 minutes, everything worked perfectly, names visible, records as before, etc.

Tree directory reduced from 15G to 1.1G.

>The way we have it configured by default, every 10,000 blocks (about 2-3 months) when the user restarts, it will compact again.

I think this is fine. Even if someone opens Bob after a long time, they wouldn't have synced and compaction will not happen (nice that it's height based, not time based).

And because #431 was merged, even the agent showed `agent: '/hsd:4.0.0/bob-wallet:0.9.0/',` :heart:

LGTM!

PR #526, opened 2022-07-01
wallet: mnemonic validation and xpriv import

### LINKED ISSUE #525: Importing seed phrase doesn't create the same wallet
`hsd` uses user-input seed phrase as-is and includes new lines. Bob doesn't strip new lines and so the wallet generated is completely different.

hsd fix: https://github.com/handshake-org/hsd/pull/731/files

Overview:
- Pasting in a seed phrase when importing a wallet accepts new lines when it shouldn't
- Wallet gets created, but with different / empty txs and balance
- Receive address and account key are different from what's expected
- Reveal seed phrase shows the standard phrase, without new lines (which is for the original wallet that the user wanted to import, but not for the wallet that has been created)

rithvikvibhu:
Closes #525.

Changes:
- **Reveal Seed**
  - Will only show seed phrase if it matches with the wallet('s key)
  - If not, will show xpriv and advice transferring everything to a new wallet
  - Will not show seed phrase when it's not available (ex: importing xpriv, see below)

- **Onboarding - Existing wallet**
  - Seed phrase input now strict and only allows letters and spaces (along with 12/24 length check)
  - New option to import from master key
    - Same flow as import with seed phrase, but textbox wants/checks xpriv
    - `walletService.importSeed` can now import both seed phrase as well as master key
    - updated i18n strings to work for both import types

code that explains what was wrong: https://gist.github.com/rithvikvibhu/48406f5b80f726e2f283644047e4acac

Bad import reveal seed phrase:
image

importing with key:
image


Falci:
Perhaps we could give some emphasis to "all your assets" and let it clear that it includes:
* domains: current and future (auctions in progress)
* coins
* HIP1 domains


About the "clear your clipboard": could Bob do that automatically? 
Just override the transfer area with "private key removed from clipboard by Bob Wallet"

rithvikvibhu:
Updated to clarify:
image

rithvikvibhu:
>About the "clear your clipboard": could Bob do that automatically?

Can think of a few problems here:
- Bob can copy new text, but can't clear clipboard history. Dunno about macos, but Windows and many Linux DEs have them enabled by default and Bob saying it has cleared it gives a false sense of security.

- There are cases where I would have plans to paste it somewhere else after Bob (for backup) and I wouldn't want an app to do it automatically for me.

PR #546, opened 2022-08-19
wallet: batch bulk actions

### LINKED ISSUE #398: Could not resolve preferred inputs
## Expected Behavior

When I click "Reveal # bids within # days for the bids to count," I should be taken to the bids page to start the process of revealing said bids.

## Current Behavior

Issue title is the result.

## Context (Environment)

I'm using Ledger, that may be the issue.

## Logs

My log file is 26MB large...I think that exceeds GitHub's limit. Is there a way to reset the log?

## Screencap

SS 2021-07-30 at 4 40 06 PM

rithvikvibhu:
Closes #398 (better handling this time) and other issues which don't have GitHub issues (described below).

### 1. Consider pending transactions for stats
When calculating balance and cards, use `getUnspentCoin` so that pending txs don't conflict with bulk actions.
Ex: Reveal All would not work if there is a bid revealed but not yet confirmed.

### 2. Find renewable names based on height, not time
This is to match hsd's condition. Also update 2 months to match hsd's 3 months

### 3. Use batching for bulk actions
Major part of this PR. It replaces registerAll, transferMany, revealAll and redeemAll with hsd's equivalent batch rpc calls. Similarly adds finalizeAll and renewAll.
Until now, all the *All functions would create large (invalid) transactions if there were many bids or names (a few hundred+). With batching, these are broken up into multiple smaller transactions.

This behaviour is slightly different with Ledger wallets:
- **Regular wallets:** Clicking "Reveal All" (or any bulk action) will repeated create and send txs until there are none left. I think this matches expected UX, and similar to UX goals in the hsd PR description
- **Ledger wallets:** Clicking it will create and send only 1 tx with as much bids/names as possible. Once each output is confirmed on the device, the button will need to be clicked again for the next batch.

>**Note:** This behavior is changed in #569. All types of wallets will always send 1 tx (the way Ledger wallets work above). Explained by in that PR.

Tested with both regular wallets and Ledger wallets.
For reviewers: be sure to run `npm install` as this currently requires https://github.com/handshake-org/hsd/pull/764
You can use this script to create a lot of opens and place bids: https://gist.github.com/rithvikvibhu/9f891202a430a50af3a1cc6375f0dfe3

h/t to @pinheadmz for adding batch to hsd!

**To do:**
- [ ] remove TEMP commit with hsd dependency change before merge (maybe replace it with commit hash)


rithvikvibhu:
GitHub build for this branch (Windows and Linux) for anyone who wants to try it out: https://github.com/rithvikvibhu/bob-wallet/actions/runs/3083707779

jvcjr:
## Solve: Method not found: sendbatch

I encountered inconsistency in my wallet history. I tried to Transfer all my TLDs into new Seedphrase wallet.
It transferred after 2 days but when I Finalize it say "Exceed Policy Weight" limit.

##

rithvikvibhu:
based on and blocked by #590. Also, while this PR doesn't have reviews, #569 is based on this and has been tested by multiple people.

enginux:
> based on and blocked by #590. Also, while this PR doesn't have reviews, #569 is based on this and has been tested by multiple people.

Go for it. Been testing it with its max N and its great!
Ref: https://discord.com/channels/822591034202521641/822591754477371443/1039817135507128370

It would be cool for Bob with multisig and batching features release!

PR #591, opened 2023-01-26
Ledger Fix script for Arch Linux and Linux users

miguelgargallo:
Ledger Fix script for Arch Linux and Linux users


rithvikvibhu:
Thanks for PR! A couple of points:
- This is not a Bob or Handshake problem, it's related to setting up ledger software to be used by _any_ wallet (kelpr, metamask, etc.) and even Ledger's official manager app.
- Having said that, it can be useful to mention it somewhere, like in the readme / install instructions. 
- This applies to Linux distros in general, not just Arch
- It's not only for development, any kind ledger connection won't work even for regular users (Bob, other wallets) who download the software and run.
- The npm install hsd-ledger is no longer required, it's already included.

Instead, how about we add a note at the end of **How to Install Bob Wallet** in the README like:

~~~md
For Linux users, if your Ledger device isn't showing up in the Ledger Live app and Bob cannot find it, some [udev rules might need to be set up](https://support.ledger.com/hc/en-us/articles/115005165269-Fix-USB-connection-issues-with-Ledger-Live?support=true) with:

```sh
wget -q -O - https://raw.githubusercontent.com/LedgerHQ/udev-rules/master/add_udev_rules.sh | sudo bash
```
~~~

PR #592, opened 2023-01-27
Chore | Add RU locale

IgorAbramov:


handshake-enthusiast:
@IgorAbramov could you update your translation with [the latest changes](https://github.com/kyokan/bob-wallet/commits/master/locales/en.json) from locale files (like new keys and keys reordering)?

PR #599, opened 2023-02-02
Minor table layout fix for "Your bids" section

handshake-enthusiast:
Otherwise a user may see "BIDDING NOW (1 BID(" (truncated) instead of "BIDDING NOW (1 BID(S))". See attached screenshots.

Just made it consistent with watching.scss ("Watching" section"):
- https://github.com/kyokan/bob-wallet/blob/03c8162a70cfc334c646b91fef9a899dd16d6038/app/pages/Watching/watching.scss#L144-L146
- https://github.com/kyokan/bob-wallet/blob/03c8162a70cfc334c646b91fef9a899dd16d6038/app/pages/Watching/watching.scss#L173-L175

Not sure why this was changed in b830e96a2e5e6ce18520cee8c9877e1b700d6908.

---

## Your Bids – Before

1 - Your Bids (before)

## Your Bids – After

2 - Your Bids (after)

## Watching

3 - Watching


rithvikvibhu:
Thanks for the PR @handshake-enthusiast! We need a more responsive table that doesn't depend on fixed widths. Until then, this should do.
Confirmed it doesn't affect any buttons in the right-most column. It does reduce the width for the name a little bit, but resizing the window will make that column grow so should be fine.

PR #606, opened 2023-02-15
Remove beta notice

handshake-enthusiast:
Is it still beta? As per versioning (2.0.0) it doesn't seem so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
 
 Bob Wallet is a [Handshake](https://handshake.org) wallet with an integrated full node.
 
-**Status**: This is beta software. As with all wallet GUIs, please use with care, and at your own risk.
-
 ## How to Install Bob Wallet
 
 Most users should use the prebuilt binaries in this repo's [releases](https://github.com/kyokan/bob-wallet/releases) page.


handshake-enthusiast:
Also does it make sense to at least mention https://github.com/kyokan/bob-wallet/releases in https://github.com/kyokan/bob-wallet/blob/master/CHANGELOG.md?

rithvikvibhu:
Version numbers don't necessarily indicate the release stage and while Bob is daily driven by many, there could still be issues. For example, in [v1](https://github.com/kyokan/bob-wallet/releases/tag/v1.0.0), we found that some wallets didn't show the correct seed phrase and the wallet users thought they had backed up was a completely different one that what was used! It was pretty rare and only affected ~2 people, but it's still 2 too many.

---

>https://github.com/kyokan/bob-wallet/blob/master/CHANGELOG.md

Wow that hasn't been updated since 0.5.0! I guess we could get it caught up with the new releases from GitHub release notes (into CHANGELOG's format). Wanna take that up @handshake-enthusiast?

handshake-enthusiast:
>while Bob is daily driven by many, there could still be issues

I can guarantee there are issues 😄 E.g. issues are still reported in Telegram and I had a number of different issues. Bugs are inevitable.

Overall, this software doesn't feel "beta".

>**Status**: This is beta software. As with all wallet GUIs, please use with care, and at your own risk.

My concern is this phrase in the beginning of the repo descripton (in README). I believe the phrase could make an impression to new users that this wallet software is unsafe. So I wanted  to make sure it's deliberately kept here and whether it's time to reconsider the project status.

handshake-enthusiast:
I can fill in CHANGELOG if you want to keep it updated. Though it may be problematic for me to adhere the format of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), particularly types of changes like Added, Fixed, etc. If you think a simple copy-paste from a release's Changelog with a reference to a corresponding release would work well – I can do it!

Releases descriptions are excellent BTW.

rithvikvibhu:
>My concern is this phrase in the beginning of the repo descripton (in README). I believe the phrase could make an impression to new users that this wallet software is unsafe.

Hm okay what can we replace that sentence with? Also, it might included there for legal reasons, I have no idea. Gonna have to make sure it's okay to change it.

---

>I can fill in CHANGELOG if you want to keep it updated. Though it may be problematic for me to adhere the format of [Keep a Changelog](https://keepachangelog.com/en/1.0.0/), particularly types of changes like Added, Fixed, etc. If you think a simple copy-paste from a release's Changelog with a reference to a corresponding release would work well – I can do it!

Instead of having different styles in the file, we can stick to your original suggestion: linking to the releases page at the top of the file. Easier and we won't have to rewrite for past (and future) notes.

handshake-enthusiast:
This is very strange, but I can't update the PR after the fork deletion even if I start with the same HEAD. I'll just close this PR to proceed in a new one.

PR #608, opened 2023-02-16
chore: restore mainWindow on MacOS

Falci:
Restore the main window from the dock (MacOS only)


app/main.js
```diff
@@ -102,6 +102,12 @@ if (isPrimaryInstance) {
       }
     });
 
+    app.on('activate', () => {
+      // On OS X it's common to re-create a window in the app when the
```
handshake-enthusiast, line 106: I believe Apple changed their OS naming from old OS X to macOS.

handshake-enthusiast (commented):
Does this address Bob Wallet app window stops showing after some time? If so this is a great change to a very annoying issue!

rithvikvibhu:
@Falci can you explain a bit on what this does or aims to fix? (I'm not a macos user :|)

From [electron docs](https://www.electronjs.org/docs/latest/api/app#event-activate-macos):
>Emitted when the application is activated. Various actions can trigger this event, such as launching the application for the first time, attempting to re-launch the application when it's already running, or clicking on the application's dock or taskbar icon.

Also, re: the change from `focus()` to `show()`, looks like `show()` is macos-only:
>app.show() macOS
>Shows application windows after they were hidden. Does not automatically focus them.

Falci:
> @Falci can you explain a bit on what this does or aims to fix?

Sure. On Mac, when we close Bob (x or cmd+q) the main window disappear but the dock shows Bob as active, as if we could restore it. But if we try to restore, nothing happens.
While there's a check to for the `window-all-closed` that prevents the app to be terminated, there was nothing there to recreate the `mainWindow` when the app is active again.


https://user-images.githubusercontent.com/1330573/220043600-e92a8f65-6fc0-4356-a3c7-15fd6f14b8db.mov

handshake-enthusiast:
I can confirm the same happens for me. Would be great to have a patch version release once it’s merged.

rithvikvibhu:
Oh cool, that makes sense, thanks!

PR #631, opened 2023-06-14
Domain Manager Multi-select and Portfolio Actions Preview

### LINKED ISSUE #570: Many feature requests
1. Multi-select and bulk actions in Domain Manager
2. View all receive addresses
3. Sidebar cleanup (portfolio rename, remove search?, etc.)
4. Refresh auction status in watchlist on new blocks
5. Alerts

See Discord chat log: https://discord.com/channels/822591034202521641/822591754477371443/1020659912767975445

### LINKED ISSUE #593: Show domains that need attention instead of number
https://i.imgur.com/yqDjlev.png
can you add what domains need attention when I get this message

rithvikvibhu:
This is a single PR for 2 features since they are somewhat related. Closes #570, Closes #593.

# Domain Manager Multi-select

Multiple domains can be selected with checkboxes.

| unselected | selected  |
|---|---|
| image | image |

- Both search filter and pagination preserve selection, so it's easier to mark domains
- The previous **Bulk Transfer** button that brought up a modal is replaced by this multi-select. With the domain selection taking place before the transfer modal, the same domain selection can be used for other actions.
- Currently, there's Transfer and Renew. Others like "Set Records" can be added similarly.
- Bulk Finalize is removed from Domain Manager:
  - Portfolio has had a Finalize card which does the exact same thing
  - The warning about ensuring Exchange names being finalized here is moved to the Finalize card preview (the next feature below) - it doesn't just warn, it compares and stops them from being finalized in bulk.

# Portfolio Actions Preview

Name related actions (Register, Renew, Transfer, Finalize) now preview when clicking the cards instead of blindly sending.

| renew, finalize | transferring, register |
|---|---|
| image | image |
| image | image |

- Reveals and Redeems don't get any preview
- It's also possible to (un)select before confirming
- This is helpful when the list is > what fits in a batch
- Checks for possible issues in preview (unregistered names can't be renewed, etc.)

# Other changes

- Bulk Transfer now handles max limits properly
- After completing a portfolio card, the pending tx shows up in the list without having to switch pages
  - it takes a short time after a tx is sent from a node to make its way to the wallet, so added a delay before refreshing pending transactions
- In the My Domains table, the "HNS Paid" column incorrectly displayed the highest bid value, not the actual value of the domain (which is the hns paid)

### DIFF
--- a/app/pages/DomainManager/BulkTransfer.js
+++ b/app/pages/DomainManager/BulkTransfer.js
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from "prop-types";
+import { consensus } from 'hsd/lib/protocol';
 import { MiniModal } from '../../components/Modal/MiniModal';
 import { Table, HeaderRow, HeaderItem, TableRow, TableItem } from '../../components/Table';
 import { connect } from 'react-redux';
@@ -10,8 +11,7 @@ import Alert from "../../components/Alert";
 import {transferMany} from "../../ducks/names";
 import {I18nContext} from "../../utils/i18n";
 
-// TODO: Research and set a good value
-const MAX_TRANSFERS_PER_BATCH = 100;
+const MAX_TRANSFERS_PER_BATCH = consensus.MAX_BLOCK_UPDATES / 6;
 
 @connect(
   (state) => ({


miguelgargallo:
It would be great to index list by date, last 30 tlds to expire, automatically renew or transfer, aka also make a default function, to those in less than X months (user can set up manually, by default 3 monhts)

app/background/wallet/bulk-actions.js
```diff
@@ -0,0 +1,167 @@
...
+
+  for (let i = 0; i < names.length; i++) {
+    const ns = names[i];
+    const { owner } = ns;
+
+    const coin = await wallet.getUnspentCoin(owner.hash, owner.index);
+
+    if (coin) {
+      if (!coin)
+        continue;
```
rithvikvibhu, line 20: found by @Falci

miguelgargallo:
the "copy" in the red warning is a link that copies text is it not understandable in UI/UX terms, a button of "Copy" in red would be great

PR #654, opened 2024-02-02
Update hsd and add soft fork changes

rithvikvibhu:
- Identify locked names and don't show auction
- update to hsd v6.1.1 ([changelog](https://github.com/handshake-org/hsd/blob/master/docs/release-notes/release-notes-6.x.md))
- update shakedex (only to remove hs-client dependency conflict)

image


app/pages/Auction/BidActionPanel/Reserved.js
```diff
@@ -1,26 +1,48 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { shell } from 'electron';
 import {
   AuctionPanel,
 } from '../../../components/AuctionPanel';
-import {I18nContext} from "../../../utils/i18n";
+import { I18nContext } from "../../../utils/i18n";
+
+const INFO_URL = 'https://heytx.substack.com/p/the-happening-event-for-handshake';
```
Falci, line 9: This could come from `locales/en.json` so other languages could link to other blog posts

